PR #43, opened 2014-12-18
Domain transfer troubleshooting article

jacegu:
I started with domain delegation issues which has been a question that has twice today on support. It's rough but I will refine it before publishing it as it was just kind of a dump of the support answer plus some structuring.

I will probably also write about #33 in here.


weppos:
Closing as per our discussion on Slack. We currently have such info in the article transfer page.

PR #46, opened 2015-01-06
Update Secondary DNS Support page for configuration instructions

jcaudle:


weppos:
Excellent starting point :+1: . I have a couple of feedback.

1. The major one is that I would probably add a small section to explain what happens in practice on a zone chance.

    In other words, the current article explains how to properly enable/disable/update the config. But then what? I would add a simple section called **Zone Changes** where we add a few lines to explain that once the secondary configuration is enabled, our system will automatically replicate the DNS records in the domain on the secondary DNS instance, that will result in the changes being replicated to all the listed secondary providers.

    Standard records will be copied as they are. Special records, such as `ALIAS`, `URL`, etc will be resolved and the result will be pushed for secondary synchronization.

    I know it seems to be obvious to a technical person, but I think a lot of less technical users may benefit from this little section.

Here's some trivial formatting feedback.

2. I'd rename the file to `secondary-dns.markdown` (no support. except in one other case, we used `_support` for other meanings). I know, this is a trivial change. ;)

3. I have been following the guideline to start generic articles (not FAQ) with the following pattern

    ```
    # Title

    ### Table of Contents {#toc}

    * TOC
    {:toc}

    ---

    Short paragraph that explains the article and or the topic.


    ## Section 1

    ...

    ## Section 2

    ...
    ```

4. Not sure if it's good or bad, but I have been following the convention to use `<label></label>` for button names, link names, rather `**bold**`. That may be useful if we decide to apply a special formatting in the future. :)

jcaudle:
One thing I was wondering is whether we should include the IP address and hostname of the AXFR name server in this document.

aeden:
The AXFR name, yes, the IP address, no (as that may change).

jacegu:
Reviewing the changes that @jcaudle did the only things I see missing given the provided feedback are:

- [ ] Provide an example for delegation changes for hosted domains as suggested by @sbastn
- [ ] Section about zone changes as suggested by @weppos

jacegu:
That's pretty much it. Proofreading of these tiny changes is more than welcome.

content/articles/secondary-dns.markdown
```diff
@@ -0,0 +1,87 @@
...
+- `ns1.dnsimple.com`
+- `ns2.dnsimple.com`
+- `ns3.dnsimple.com`
+- `ns4.dnsimple.com`
+- `ns1.secondary.com`
+- `ns2.secondary.com`
+
+Remember that you will need to revert this step if you ever disable Secondary DNS for your domain.
+
+</warning>
```
jcaudle: This warning section seems a little big now. Should we move the section about hosted domains to its own proper section?

content/articles/secondary-dns.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+Before setting up AXFR with our systems it will be necessary for you to choose, activate, and gather information for the secondary DNS provider you will be using. Once you have a secondary provider configured, you should continue with this guide.
+
+<warning>
+It is possible to configure Secondary DNS on all domains whether they delegate to us or not, but if you have pointed delegation elsewhere you will need to [point the domain to DNSimple](http://support.dnsimple.com/articles/pointing-domain-to-dnsimple/) before the configuration will take effect.
```
jcaudle: I forgot to use a relative link here.

martinisoft:
Copy wise, this looks good to me. :+1: to :ship:

content/articles/secondary-dns.markdown
jacegu: Thought about it but couldn't come up with a good section title. It feels to me that maybe these warning can be split up added to the *Enable* and *Disable* sections under a subsection that describes what else to do if your domain is hosted.

However I think we rather keep it this way for now and see how it goes for now.

weppos: > This warning section seems a little big now. Should we move the section about hosted domains to its own proper section?

I agree. Same applies to me for the **What happens once Secondary DNS is enabled?** info section.

> Thought about it but couldn't come up with a good section title. It feels to me that maybe these warning can be split up added to the Enable and Disable sections under a subsection that describes what else to do if your domain is hosted.

What about **Configure delegation**?

PS. As a side note, I have been following the [Heroku DevCenter](https://devcenter.heroku.com/articles/writing) guidelines for all my articles in support. We should use "Enable Secondary DNS" rather than "Enabling Secondary DNS" for section titles.

The section **What happens once Secondary DNS is enabled?** can be called **Zone Changes** or **Behind the scenes**

weppos: Actually, for the warning, we can also call it "Getting started" and move the entire initial paragraphs from "Before setting" (including the warning, without the warning formatting) into that section.

Issue #17, opened 2014-04-12
Article about invoices


jacegu:
Covering where to find them, how to change information, etc..

jacegu:
This is widely covered in https://support.dnsimple.com/articles/account-payment-history/

PR #47, opened 2015-01-13
Article about billing information on invoices

jacegu:
Related to #17 

Covers how billing information is displayed on invoices and how to change it.


jacegu:
Ready for proofreading

PR #49, opened 2015-01-22
Add information about special WHOIS policies

weppos:


jcaudle:
Looks good to me. I say :shipit:. I think there may be special policies for other TLDs, but I'll have to review some of the errors we've gotten from domains to be able to say which ones need special intervention.

PR #51, opened 2015-02-04
Old instructions for Zone export

sbastn:
This fixes the old screenshot and matches the copy to the correct links.


weppos:
:+1:

PR #55, opened 2015-03-05
Provide detailed instructions on how to configure secondary DNS with known 3rd party providers

sbastn:
This PR contains detailed instructions on how to set up secondary DNS with our known providers

- [x] DNSMadeEasy
- [x] EasyDNS


aeden:
Post images should go in `/files`, not `/assets/images`.

I've merged https://github.com/aetrion/dnsimple-support/pull/54 that moves all externally hosted images into that directory. This PR should be updated to do the same.

content/articles/secondary-dns-provider-dns-made-easy.markdown
```diff
@@ -0,0 +1,102 @@
+---
+title: Secondary DNS with DNSMadeEasy
+excerpt:
```
jcaudle: How does "Secondary DNS can be complicated to set up. We've simplified things with provider specific settings for DNSMadeEasy." sound?

content/articles/secondary-dns-provider-easy-dns.markdown
```diff
@@ -0,0 +1,100 @@
+---
+title: Secondary DNS with EasyDNS
+excerpt:
```
jcaudle: This could probably be the same as whatever we decide on with DNSMadeEasy, just with the provider name swapped out.

PR #58, opened 2015-03-13
Set up XML sitemap

jcaudle:
This will generate a sitemap every time we compile the site for publication.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'nanoc', '~> 3.7.0'
+gem 'builder'
 gem 'kramdown'
 gem 'mime-types'
 gem 'nokogiri'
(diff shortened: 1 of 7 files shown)


weppos:
:+1: You may want to add the location in the [`robots.txt` file](http://www.sitemaps.org/protocol.html#submit_robots).

jcaudle:
We don't seem to have a robots.txt right now. I don't believe there's anything we want to disallow, correct?

martinisoft:
@jcaudle Still good to have to let the bots know what is really crawl-able.

weppos:
@jcaudle you can create a blank `robots.txt` with only the `Sitemap:` declaration. Search Engines will accept it.

jcaudle:
Yep. I can get one added here and in aetrion/dnsimple-developer#24.

weppos:
:metal:

PR #60, opened 2015-04-17
Add note about Secondary DNS

jacegu:
A customer was confused about this bit as he was considering his setup incorrect despite having Secondary DNS enabled:

> 2. When I run `dig NS smockle.com +short` I see “c.ns.buddyns.com.” in addition to the four DNSimple name servers. This page: https://support.dnsimple.com/articles/domain-resolution-issues/ lists that as a problem.

I have added a small note clarifying that it doesn't apply to domains with Secondary DNS.


jcaudle:
:shipit:

PR #71, opened 2015-07-03
Two-factor documentation

weppos:
This PR adds the documentation that explains what is and how to enable/disable two-factor authentication.

/cc @aetrion/dnsimple for review


aeden:
:shipit:

PR #75, opened 2015-08-12
Update migration docs to show 2FA customers are forced to migrate

jcaudle:
The method of migrating away from Authy's 2FA is now obsolete due to the forced login for people who hadn't migrated already. This clears up some of what was causing confusion.


weppos:
:+1:

PR #76, opened 2015-08-19
Article to upgrade single-hostname SSL certificate

weppos:
/cc @aetrion/dnsimple for review


arnamak:
Everything looks good to me :+1:, however you might consider swapping the second paragraph for additional clarity. Perhaps something like: 

"However, if you decide a wildcard better fits your needs within 30 days of purchasing a single-hostname SSL certificate, we can refund your initial order so you are only charged for the wildcard SSL certificate."

weppos:
@arnamak feel free to edit it directly

PR #78, opened 2015-08-27
Add an article about account payment history

jcaudle:
This article includes instructions on how to view payment history and
retry failed invoices.

Any comments on the article from @aetrion/dnsimple are greatly appreciated.


jacegu:
I would add a note that points out that invoices are periodically retried by our system.

jacegu:
The other thing that we may want to add is an explanation of what "Dunned" invoices are and why they cannot be retried by our customers.

At this point a section that explains different invoice states is starting to make sense. I'll be happy to jump in and add it.

Issue #81, opened 2015-10-01
"Sharing Domain Management with Another User" points to a blog post from 2011


sbastn:
Update https://support.dnsimple.com/articles/sharing-domain/ with new instructions

PR #82, opened 2015-10-01
Sharing domain management

sbastn:
This PR updates the `/articles/sharing-domain/` page removes the link to an old blog post from 2011 and updates the instructions.

Based on #81


sbastn:
Ready for your :eyes:

weppos:
In general, I feel like we may want to ship more structured articles (list/add/remove sections, etc) but this PR is definitely better than what we have, hence I'd say :shipit:

PR #22, opened 2015-10-08
Article about ICANN domain validation

aeden:
We should add a domain that explains what is the ICANN domain validation and what will happen to the domain if it's not validated.

We should mention the name servers being changed to the following entries and the domain state being changed to suspended.

```
Name Server:DNS1.NAME-SERVICES.COM
Name Server:DNS2.NAME-SERVICES.COM
Name Server:DNS3.NAME-SERVICES.COM
Name Server:DNS4.NAME-SERVICES.COM
Name Server:DNS5.NAME-SERVICES.COM
```


content/articles/icann-domain-validation.markdown
```diff
@@ -0,0 +1,46 @@
...
+
+ICANN, the organization that oversees domain name registrations for most top-level domains, requires validation of your registrant email address whenever a new domain is registered, or your registrant email address or name is changed. Failure to validate your registrant email address results in suspension of the domain name after 15 days of non-compliance.
+
+When you make a change to your registrant's email address or name, you will receive an email similar to the following:
+
+![ICANN Verification Email](/files/icann-verification-email.png)
+
+This email will be sent from a `@dnsimple.com` email address and will include a link similar to the following:
+
+`http://raa.name-services.com/raaverification/verification.aspx?VerificationCode=A8E3763E-EE70-42DB-A654-20BF560300A00`
```
jcaudle: Should we remove the `VerificationCode`?

jcaudle:
Besides my question about the `VerificationCode` (which is now outdated because of 7d1a3ff ¯\\\_(ツ)_/¯), I think this is ready to go.

content/articles/icann-domain-validation.markdown
aeden: I already changed the text... I can change it further if you want (to all zeros for example).

weppos: As far as the text is "random", I'd prefer it over the zeros.

jcaudle: Yeah, I'd prefer a randomized code to all zeros. I say leave it and :shipit:.

Issue #19, opened 2014-05-09
Article about importing zone files


jacegu:

jacegu:
Duplicate of #10

Issue #10, opened 2014-04-03
Article about importing records


jacegu:
Using:

- Domain zone file
- API ?

PR #86, opened 2015-10-08
Zone import article

sbastn:
This PR adds a new zone import article. Based on #19.


weppos:
@sbastn I have the feeling that @jacegu may have come with a more comprehensive article in #101. Do you confirm we can use #101?

sbastn:
Yep. Let's use #101

jacegu:
So sorry about this man. Didn't notice there was a duplicate issue and we started working on the same thing.

sbastn:
@jacegu no worries, we had 2 issues for this #10 and #19. I assigned #19 to myself without understanding that #10 was a kinda of duplicate.

Anyway, yours is better. You win, I win, we all win!

PR #90, opened 2015-10-08
Improve drop catching article

aeden:
This PR improves the drop catching article by explaining what it is, why we do not allow it, and what we will do if we see it.

### DIFF
--- a/content/articles/drop-catching.markdown
+++ b/content/articles/drop-catching.markdown
@@ -1,10 +1,26 @@
 ---
-title: Can I use DNSimple for Drop-catching?
-excerpt: Information about DNSimple drop-catching policy.
+title: Can I use DNSimple for Drop catching?
+excerpt: Information about DNSimple drop catching policy.
 categories:
 - DNSimple
+- Domain Registration
+- API
 ---
 
-# Can I use DNSimple for Drop-catching?
+# Can I use DNSimple for drop catching?
 
-No, DNSimple may not be used for drop-catching domains at this time.
+No, you may not use DNSimple for drop catching domains at this time.
+
+## What is drop catching?
+
+From [Wikipedia](https://en.wikipedia.org/wiki/Domain_drop_catching):
+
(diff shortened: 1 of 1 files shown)


content/articles/drop-catching.markdown
```diff
@@ -1,10 +1,26 @@
...
+
+## What is drop catching?
+
+From [Wikipedia](https://en.wikipedia.org/wiki/Domain_drop_catching):
+
+> Domain drop catching, also known as domain sniping, is the practice of registering a domain name once registration has lapsed, immediately after expiration.
+
+## Why do you not allow drop catching?
+
+Our upstream domain registration provider limits the number of checks we are allowed to execute. If you attempt to engage in drop catching then this will impact all of our customers.
```
jcaudle: Instead of just "number of checks" should we say "number of registration checks"? I don't know if "check" is a term that's commonly used in this scenario, so feel free to disregard if it is.

jcaudle:
:+1: :sheep: :it:

content/articles/drop-catching.markdown
```diff
@@ -1,10 +1,26 @@
 ---
-title: Can I use DNSimple for Drop-catching?
-excerpt: Information about DNSimple drop-catching policy.
+title: Can I use DNSimple for Drop catching?
+excerpt: Information about DNSimple drop catching policy.
 categories:
 - DNSimple
+- Domain Registration
+- API
```
weppos: FYI, each article so far was assigned to exactly one category. It is not a strict requirement (it was in previous systems as of platform requirements) but in order to limit issues I decided to stick with this implicit rule.

Do you feel like it's better to assign multiple categories? As a side note, I don't even know if the platform supports them  as I never tried.

jcaudle: It looks quite nice.

screen shot 2015-10-08 at 10 56 13 am

I like the idea of making things available in whatever categories make sense as it can lead to better discoverability, I think. Even I've had trouble tracking down something that I was looking for as I thought it could fit in multiple categories and it took some time to find the right one. Often I just resort to the search bar when I have trouble figuring out what category to look in.

PR #92, opened 2015-10-08
Add reasons for suspension and improved info on how to contact us.

aeden:
Improvements include listing of reasons for suspension as well as an extended explanation about how to contact us and what to include in the support request.


sbastn:
:sheep:

Issue #93, opened 2015-10-08
Make it easier to search from the landing page


sbastn:
Based @Barbamama's notes:
> To find the answers to your questions, either enter text into the search box above and search for articles or click on one of the topics to the right. I am already on the help page so it makes sense to me to have a search box right there and not a long sentence.

Show the search box right on the landing page and adjust the copy

PR #95, opened 2015-10-08
Landing page

sbastn:
Adding a search box on the landing and reorganized the copy. Based on #93

### Before
screen shot 2015-10-08 at 1 10 37 pm

### After
screen shot 2015-10-08 at 1 04 40 pm


aeden:
:100:

Barbamama:
It's way better and very clear, thank you Sebastien

aeden:
One quick question: I keep going back and forth amongst: 

1. always providing a mailto to support
2. linking to the contact page on the DNSimple site
3. having both the email address and a link to the contact page

Any preference?

Barbamama:
Option 1 instead of going back to an other page again, I would rather be able to send my request.  Make it at simple and fast as possible.

sbastn:
Option 1 here as well. You are already on the support documentation site.

sbastn:
I refactored the search into a partial to match the same style:

### Before
screen shot 2015-10-08 at 1 21 33 pm

### After
screen shot 2015-10-08 at 1 19 12 pm

Consistency, consistency, consistency!!!

Ready for your :eyes:

aeden:
It sounds like the home page should be switched to pure HTML. Do you want to do that as part of this PR?

weppos:
@Barbamama @aeden 

> linking to the contact page on the DNSimple site

I vote for this option. And there is a reason I mentioned in the past (the reason why most of the articles/links in our site point to the support page).

Using option (1) has the following (not irrelevant) issues:

1. The user must have an email account configured on the device it's using or be able to access the web UI. Although this is generally true, there are 3 real cases where it's not so common:
    - mobile devices
    - public devices (internet cafe)
    - corporate/firewall/specific environements
1. Clicking on a mailto link in general causes the underlying system/device to fire off the email client or the handler registered for mailto. Some users (including myself) find it annoying because they may have different emails (private vs corposate) configured on different tools and I want to be able to use the one I want (whereas the handler can't allow multiple configuration).

The advantage to point to a page is that the page allows the user to simply type the email in a form without the need to have a mail client or tool configured on the device. It still shows the full email, so that the user can also decide to send us an email directly using his/her favourite handler.

sbastn:
Converted to HTML and linking to the contact page on the app.

Ready to merge?

jcaudle:
:clap: This looks really nice @sbastn.

PR #98, opened 2015-10-08
Adding information for URL rewrites

therobot:
Based on a support request, clarify that complete URL rewrites can't be done with our URL redirect record.


weppos:
I agree with the concept, but I believe it should be rephrased. The term **URL Rewriting** is generally associated with the idea of rewriting ID-based URLs into "speaking URLs". For example, in WordPress and CMS in general you configure the URL rewriting to convert

    /article.php?id=1

to

    /articles/why-dnsimple-is-cool

In fact, it's not even completely clear to me the meaning of the added paragraph. Is it related to the fact that we don't allow 1-to-1 page to page redirects (exact matches) or that we don't support redirect matches (aka `(.*).example.com` to `whatever.com/$1`)?

I believe an example would also help to clarify.

PR #106, opened 2015-10-08
Article cloudflare

weppos:
Add information about enabling/disabling and explain the difference between the two integration methods.

### DIFF
--- a/content/articles/cloudflare-service.markdown
+++ b/content/articles/cloudflare-service.markdown
@@ -28,13 +28,13 @@ CloudFlare has two integration methods:
 
 Direct integration requires to change the name servers for your domain to point to CloudFlare name servers.
 
-If your domain is registered with DNSimple, follow these instructions to [configure the name servers for a domain](/articles/setting-name-servers/). If the domain is registered elsewhere, you will need to follow the instructions of your existing domain registrar to point the domain to CloudFlare. In both cases, once the domain is pointint to CloudFlare, you will need to manage the DNS records in CloudFlare. Any DNS change in your account for that domain will no longer have any effect, including changing DNS records, enabling/disabling one-click services and enabling/disabling email forwards.
+If your domain is registered with DNSimple, follow these instructions to [configure the name servers for a domain](/articles/setting-name-servers/). If the domain is registered elsewhere, you will need to follow the instructions of your existing domain registrar to point the domain to CloudFlare. In both cases, once the domain is pointing to CloudFlare, you will need to manage the DNS records in CloudFlare. Any DNS change in your DNSimple account for that domain will no longer have any effect, including changing DNS records, enabling/disabling one-click services and enabling/disabling email forwards.
 
-Furthermore, the direct integration doesn't allow to selectively delegate only certain DNS records to CloudFlare. If you want to use CloudFlare only for a specific DNS record (e.g. for the main website) and use DNSimple for all the other records (e.g. MX email configurations), then use the one-click DNSimple integration.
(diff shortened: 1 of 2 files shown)


content/articles/cloudflare-service.markdown
```diff
@@ -16,10 +24,45 @@ CloudFlare has two integration methods:
...
+
+
+## Universal SSL limitations
+
+[CloudFlare Universal SSL](https://www.cloudflare.com/ssl) is currently made available only if you use Cloudflare directly (first integration method). It's not possible to use CloudFlare Universal SSL if you enable Cloudflare using DNSimple one-click service.
+
+
+## Enabling CloudFlare service {#enable}
+
+From the domain page, switch to the <label>Services</label> tab and click to <label>Add or edit service</label>. Find the **CloudFlare** service, click to <label>Add</label> and follow the instructions
```
weppos: I'd probably convert this section (as well the Google apps equivalent) to steps at some point.

weppos:
Ready for review.

content/articles/cloudflare-service.markdown
```diff
@@ -16,10 +24,45 @@ CloudFlare has two integration methods:
...
+
+
 ## DNS limitations
 
-CloudFlare caches your site content at locations across the globe. It uses DNS to route your requests to one of these locations depending on various factors. To support this technology within DNS limitations it is best to use a host name (i.e. www.example.com) rather than your domain apex (i.e. example.com). CNAME records at a domain apex are disallowed or will cause your DNS respond incorrectly. You may wish to use DNSimple's URL redirection record to forward requests from your domain apex to "www" for instance.
+CloudFlare caches your site content at locations across the globe. It uses DNS to route your requests to one of these locations depending on various factors. To support this technology within DNS limitations it is best to use a host name (i.e. www.example.com) rather than your domain apex (i.e. example.com). CNAME records at a domain apex are disallowed or will cause your DNS respond incorrectly. You may wish to use DNSimple's URL redirection record to forward requests from your domain apex to "www" for instance.
```
jcaudle: I'm a little confused by this sentence:

> To support this technology within DNS limitations it is best to use a host name (i.e. www.example.com) rather than your domain apex (i.e. example.com).

I think we definitely need a comma after `limitations`, but I'm not sure I completely understand what we mean by `DNS limitations`.

content/articles/cloudflare-service.markdown
```diff
@@ -16,10 +24,45 @@ CloudFlare has two integration methods:
...
+## Enabling CloudFlare service {#enable}
+
+From the domain page, switch to the <label>Services</label> tab and click to <label>Add or edit service</label>. Find the **CloudFlare** service, click to <label>Add</label> and follow the instructions
+
+
+## Disabling ClouldFlare service {#disable}
+
+<warning>
+A known bug currently prevents the possibility to disable the CloudFlare service from the UI. Please [contact us](https://dnsimple.com/contact) if you want to disable the service.
+</warning>
```
jcaudle: Should we include a note about how to workaround this issue? It's possible to workaround, but it's rather cumbersome.

jcaudle:
I made a couple tweaks to the copy and left a few questions, but I think this is ready to :ship: since the biggest question I had was already there previously.

content/articles/cloudflare-service.markdown
weppos: I think for now we can manage it manually until we eventually get it fixed.

content/articles/cloudflare-service.markdown
weppos: @aeden may help better as I think he wrote this section in the past.

aeden: DNS limitations in this case means the use of a CNAME on the root (which is not allowed). In fact, the real issue is that the current integration does not support the use of the ALIAS record in conjunction with CloudFlare.

PR #109, opened 2015-10-08
Add tests and configure Travis

weppos:
This PR adds some simple content tests (as for the blog) and setup the Travis integration.

### DIFF
--- a/_test/content_test.rb
+++ b/_test/content_test.rb
@@ -1,0 +1,33 @@
+require 'test_helper'
+
+describe "Content" do
+
+  BannedCharacters = [
+    "\u201C", # U+201C “
+    "\u201D", # U+201D ”
+    "\u201E", # U+201E „
+    "\u2018", # U+2018 ‘
+    "\u2019", # U+2019 ’
+    "\u201A", # U+201A ‚
+    "\u201B", # U+201B ‛
+  ]
+
+  it "is sucessful" do
+    regexp   = /[#{BannedCharacters.join}]/
+    affected = []
+
+    Dir["content/*"].each do |path| 
+      next unless File.file?(path)
+      File.readlines(path).each do |line|
+        affected << [path, line] if regexp.match(line)
+      end
+    end
+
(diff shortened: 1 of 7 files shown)


jcaudle:
This is great. I say :shipit:

I wonder if we might want to think about putting some sort of spell checker in place, but I don't know how well it'd work.

weppos:
I'm open to suggestions. For now I didn't add anything fancy, the main purpose was to avoid weird chars and have a test suite I could connect with Travis.

weppos:
One possible spec could be an attempt to generate a full site. This would be particularly helpful if [we decide to use Travis for deploys](http://www.paperplanes.de/2013/8/13/deploying-your-jekyll-blog-to-s3-with-travis-ci.html).

_test/content_test.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  BannedCharacters = [
+    "\u201C", # U+201C “
+    "\u201D", # U+201D ”
+    "\u201E", # U+201E „
+    "\u2018", # U+2018 ‘
+    "\u2019", # U+2019 ’
+    "\u201A", # U+201A ‚
+    "\u201B", # U+201B ‛
+  ]
```
yob: I found my way here via the dnsimple blog, curious about the kinds of tests that might apply to a documentation repo.

Why are these specific characters banned?

weppos: There are a couple of reasons. They are not ASCII characters and we had cases in the past where Markdown processors or tools we used crashed because they were not able to deal with them by default (especially before Ruby 2.0 when input was not considered UTF-8 by default).

Moreover, these characters have valid ASCII alternatives such as `"` and `'` and most of the time they are injected in posts or articles just because someone is copying and pasting from a Word document. In other words, it never happened to us that the use of these chars was intentional, and their use always leaded to troubles or make it more complex to grep/search patterns in our code.

This test it's a way for us to prevent the use of these chars in favor of the ASCII alternatives.

yob: Thanks for clarifying!

PR #110, opened 2015-10-08
URL redirections

therobot:
This PR attempts to add more information on what is possible to do and what not with URL records. It also provides examples.


content/articles/url-record.markdown
```diff
@@ -25,6 +25,15 @@ Enter the hostname you would like to redirect from, and the URL to redirect to.
 
 Right now it is not possible for our redirection service to handle SSL requests for your domain names. Take a look at the article [Can a URL record redirect requests over SSL?](/articles/url-redirect-ssl) for additional information and an explanation of alternatives.
 
-## URL Rewrites
+## URL to URL redirects
+
+URL to URL redirects can't be done with our URL record. Only domains or subdomains can be redirected to complete URLs. 
+
+Example: foo.com/blog to bar.com/blog
```
jcaudle: Maybe add a note that the desired behavior (that's not possible with DNS) is to only redirect foo.com/blog, but not foo.com or foo.com/something-else

therobot: Good one jcaudle

content/articles/url-record.markdown
```diff
@@ -25,6 +25,15 @@ Enter the hostname you would like to redirect from, and the URL to redirect to.
...
+
+## Wildcard subdomain to URL redirects
+
+Is not possible to redirect wildcard subdomains to URLs
+
+Example: *.foo.com to bar.com/$1
```
weppos: This line is partially incorrect, as we do support wildcard redirects but we don't support redirects based on matches. I'd rephrase the section to

```
## Match-based URL redirects

You currently cannot configure a redirect target that includes a back-reference (match) from the redirect source. For example, the following redirects will not work:

    *.foo.com to bar.com/$1
    *.foo.com to $1.bar.com
```

I also indented the example to appear as code.

weppos: We can also add a section for wildcard.

```
## Wildcard URL redirects

You can configure a wildcard redirect using the same conventions of a DNS wildcard record. However, the target cannot contain any wildcard reference.

    # Valid redirects
    *.foo.com to bar.com
    *.foo.com to bar.com/path

    # Invalid redirects
    *.foo.com to *.bar.com

The wildcard pattern can only be on the left outermost level.

    # Valid redirects
    *.foo.com to bar.com

    # Invalid redirects
    foo.*.com to bar.com
```

therobot: Thanks for the additions @weppos

content/articles/url-record.markdown
```diff
@@ -24,3 +24,36 @@ Enter the hostname you would like to redirect from, and the URL to redirect to.
 ## Redirecting over SSL
 
 Right now it is not possible for our redirection service to handle SSL requests for your domain names. Take a look at the article [Can a URL record redirect requests over SSL?](/articles/url-redirect-ssl) for additional information and an explanation of alternatives.
+
+## URL to URL redirects
+
+URL to URL redirects can't be done with our URL record. Only domains or subdomains can be redirected to complete URLs. The following redirect will not work:
+
+    foo.com/blog to bar.com/blog
```
jcaudle: I don't think this example fully explains what we're trying to say here. What about this:

> The following redirect will not work:
>
>     foo.com/blog/ to bar.com/blog
>
> because DNS records cannot be added to URLs. However, this would work:
>
>     blog.foo.com to bar.com/blog
>
> since a URL record could be added to blog.foo.com.

content/articles/url-record.markdown
```diff
@@ -24,3 +24,36 @@ Enter the hostname you would like to redirect from, and the URL to redirect to.
...
+## Wildcard URL redirects
+
+You can configure a wildcard redirect using the same conventions of a DNS wildcard record. However, the target cannot contain any wildcard reference.
+
+    # Valid redirects
+    *.foo.com to bar.com
+    *.foo.com to bar.com/path
+
+    # Invalid redirects
+    *.foo.com to *.bar.com
```
jcaudle: Rather than labeling this as an invalid redirect, perhaps we should note that it would actually redirect to the hostname `*.bar.com` rather than passing the captured subdomain through to the target.

content/articles/url-record.markdown
```diff
@@ -24,3 +24,36 @@ Enter the hostname you would like to redirect from, and the URL to redirect to.
...
+
+The wildcard pattern can only be on the left outermost level.
+
+    # Valid redirects
+    *.foo.com to bar.com
+
+    # Invalid redirects
+    foo.*.com to bar.com
```
jcaudle: It might be better to use an example like `foo.*.bar.com` since no one has access to the full `com` zone from DNSimple. :wink:

content/articles/url-record.markdown
```diff
@@ -24,3 +24,36 @@ Enter the hostname you would like to redirect from, and the URL to redirect to.
...
+
+## Match-based URL redirects
+
+You currently cannot configure a redirect target that includes a back-reference (match) from the redirect source. For example, the following redirects will not work:
```
jcaudle: Again, this note should probably say that they will not work as expected.

jcaudle:
Nice work on this one @therobot. I think the examples could use a little clarification, but let me know if you have any questions.

content/articles/url-record.markdown
therobot: @jcaudle the hostname *.bar.com is invalid as per the RFC https://tools.ietf.org/html/rfc952

aeden: @therobot the wildcard is simply and indicator used to match whatever subdomain was provided.

content/articles/url-record.markdown
therobot: @jcaudle I did not get this one, sorry.

weppos:
@therobot this PR is almost 2 months old. Would you mind to apply the remaining changes (if any) and merge it?

content/articles/url-record.markdown
therobot: @jcaudle can you help me on this one?

jcaudle: Oh, what I was saying is that `foo.*.com` is not a record any of our customers will be able to set, but someone who owns `bar.com` could attempt to set a redirect from `foo.*.bar.com` to go to something else (`baz.com`) for example. Does that make more sense?

Issue #111, opened 2015-10-08
Inconsistent font-size for steps


meeunier:
ex.: http://support.dnsimple.com/articles/account-activation/

`class="section-steps"` should have same font size as `<p>`

weppos:
@meeunier FYI I decreased it on purpose (I believe it was 90%) to make it more compact and different than the body of the page. That said, I'll leave the final word to @sbastn

sbastn:
Could we do something like this?
<img width="765" alt="screen shot 2015-10-08 at 9 12 03 pm" src="https://cloud.githubusercontent.com/assets/80610/10377027/4c58f834-6e01-11e5-88fe-9d00b692f161.png">

PR #112, opened 2015-10-08
Styling the "steps" section.

sbastn:
This PR fixes the styles section for "steps". Based on #111.

### Before
screen shot 2015-10-08 at 9 20 58 pm
### Now
screen shot 2015-10-08 at 9 19 24 pm

### DIFF
--- a/content/assets/css/_application.scss
+++ b/content/assets/css/_application.scss
@@ -123,7 +123,7 @@ h2 {
   background: lighten($accent-color, 50%);
   color: lighten($accent-color, 10%);
 
-  p {
+  > p {
     &:first-child {
       &:before {
         margin-right: 10px;
--- a/content/assets/css/_application.scss
+++ b/content/assets/css/_application.scss
@@ -120,8 +120,12 @@ h2 {
 .section-steps {
   margin-bottom: $base-unit;
   padding: $base-unit / 2;
-  background: lighten($accent-color, 50%);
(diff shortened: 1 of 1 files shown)


weppos:
@sbastn in several cases the steps have images on them. Just make sure you are happy with the result of images on the light blue background.

sbastn:
@weppos just noticed. What do you think? I might have to sleep on this one ;)
screen shot 2015-10-08 at 9 24 50 pm

jcaudle:
Should we go even further and add the ability to use a `<steps>` element instead of the `div` with the special class?

weppos:
@sbastn honestly I'm not thrilled about it, but I have the feeling you've got the same feeling :smiley:

jcaudle:
For what it's worth, I think the screenshot looks fine on the blue background. It even stands out a bit better than against a white background.

weppos:
Specifically, what doesn't convince me is the blue text on a light blue background that is almost identical to the link style and is also confusing with the link in the screenshot.

sbastn:
I have removed the background, @weppos was right, it can be confusing since it's the same as the notes and added a small padding and borders. Here is how it looks:

### Without image
screen shot 2015-10-09 at 8 43 04 am


### With image
screen shot 2015-10-09 at 8 43 13 am

sbastn:
I am going to merge this because I am annoyed by what we have out there and it's not even 9am.

PR #48, opened 2015-10-08
Secondary DNS FAQ

jcaudle:
We may want to consider to add a FAQ section to the secondary DNS article, as we did here.
https://github.com/aetrion/dnsimple-support/blob/master/content/articles/github-student-pack.markdown

Here's a possible list of FAQ

- [x] How do I check that secondary DNS is setup correctly?
- [x] How can I test that secondary DNS is working?
- [x] Is the order of the name servers relevant? (I was considering to promote this a general faq, with a dedicated article)

We also have a possible list of Caveats

-  [x] **Registered domain delegation**

   When you enable the secondary DNS for a domain registered with us, at the time being we automatically update the delegation to include the secondary DNS. That means, we replace the delegation if the domain is registered with us and delegated externally.

### DIFF
--- a/content/articles/secondary-dns.markdown
+++ b/content/articles/secondary-dns.markdown
@@ -1,6 +1,6 @@
 ---
 title: Secondary DNS
-excerpt:
+excerpt: This page provides information about secondary DNS configuration with DNSimple.
 categories:
 - Advanced DNS
 ---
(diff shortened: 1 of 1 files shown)


weppos:
:+1: :shipit: 

I'm merging it so that it will count in the report.

PR #116, opened 2015-10-27
Note that we may not be configured as a secondary

jcaudle:
We've been getting more requests about this and it seemed worth updating this page.


weppos:
:+1:

PR #118, opened 2015-11-05
Article about transferring UK and CO.UK domains to DNSimple.

aeden:
I have added this article because the process is unique
and we receive quite a few support requests about it.

### DIFF
--- a/content/articles/transferring-uk-domain.markdown
+++ b/content/articles/transferring-uk-domain.markdown
@@ -1,0 +1,74 @@
+---
+title: Transferring a UK or CO.UK Domain to DNSimple
+excerpt: How to transfer your registered UK or CO.UK domain to DNSimple.
+categories:
+- Domain Transfers
+---
+
+# Transferring a UK or CO.UK Domain to DNSimple
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+UK and CO.UK domains have a slightly different transfer process than most other domain registrations. This document explains some of those differences.
+
+## Before You Start
+
+<warning>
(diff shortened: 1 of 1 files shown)


arnamak:
Looks good to me, I'd maybe just suggest adding (or removing) the quotes from one of the "Press Transfer Domain" list items.

content/articles/transferring-uk-domain.markdown
```diff
@@ -0,0 +1,74 @@
...
+
+## Approve the transfer
+
+To complete the transfer of a UK or CO.UK domain, you must change the domain's IPS TAG at your current domain registrar. Your registrar may allow you to do this through their web site, or you may need to contact them via email or phone.
+
+Request the change of the IPS TAG to `ENOM` (this is case-sensitive, so make sure it is in all capital letters).
+
+Once you've authorized the transfer, you may have to **wait up to 48 hours for the transfer to complete.**
+
+<callout>
```
jcaudle: I didn't know about `<callout>`. :+1:

content/articles/transferring-uk-domain.markdown
```diff
@@ -0,0 +1,74 @@
...
+We will only charge your credit card for the transfer once it has completed.
+</callout>
+
+## Transfer status
+
+You can go to your domain list at any time to check on the status of pending transfers. Click on "transfer in process" to view detailed information about the transfer request.
```
jcaudle: It may be better to use `<label>` here around "transfer in process".

weppos:
We already have an article about tranferring a UK domain: http://support.dnsimple.com/articles/domains-uk/

I suggest to expand/update it, rather than publishing a new one. Otherwise we will end up duplicating the articles.

aeden:
The reason I created a new article is because I wanted a *specific* long-form article for the steps in transferring UK domains, whereas the article you mentioned is about all things UK (registering, transferring, etc). In other words, this was intentional.

weppos:
Then I suggest to at least remove the transfer info in the big article and link to the specific one, so that we have some crosslinking in place.

aeden:
Done.

PR #123, opened 2015-11-05
Search on site

sbastn:
This PR enables to search on the support site instead of jumping to google.


jcaudle:
I was able to get this to work locally, but it seemed pretty slow (page loads are really slow for me in development too though). Do you have any guesses as to what it'll look like in production?

PR #120, opened 2015-11-05
Article on what happens when a domain expires.

therobot:
This is basically a ripoff of @meeunier's answer to a [one particular ) support request](https://dnsimple.groovehq.com/groove_client/mailboxes/6691/filters/61705/tickets/11004081)

Feel free to add suggestions on grammar or content.

PR #125, opened 2015-11-06
What happens when a domain expires

meeunier:
Describes what happens when a domain expires.


meeunier:
Draft – I've edited it from enom's http://www.enom.com/kb/kb/kb_0402-what-happens-domain-expires.htm?Highlight=expires

Will further tweak.

weppos:
Is this a duplicate of #120? If so, which one should we keep?

PS. @therobot @meeunier in order to qualify for the OMG day, remember we have 24 hours for creating and merging the content. :stuck_out_tongue_winking_eye:

meeunier:
fuckit

weppos:
Generally, the author merges the article after a positive feedback. It generally happens after a few hours or by the end of the day in a specific session.

What I generally do is that I spend the first morning producing content, I then publish it for reviews and I spend a little bit of time reviewing other PRs. Other users are generally following a similar approach, therefore I can expect my content to be reviewed at some point during my afternoon and I merge it before evening.

:us: people are also generally providing early feedback during their day and after that they produce PRs. Likewise, they tend to get feedback by other :us: people and/or they may receive feedback by European people in the early morning of the next day.

As a result, the cycle is that we generally tend to merge all European contributions late in the evening and US contributions late in their evening (and or early in European morning).

I'd say it's responsibility of each author to make sure he/she gets a review in time to merge the article. Sometimes I don't get feedback, what I do is that I tend to ping people directly so that I can get a feedback before leaving.

aeden:
This article references other articles without links. Please link them.

meeunier:
too slow @jacobo , I win.

content/articles/what-happens-when-a-domain-expires.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Whois Privacy may cause transfer approval emails to not be delivered
+excerpt: Whois Privacy services often do not properly deliver transfer approval emails, and thus should be disabled before requesting a transfer.
```
weppos: I think this title and excerpt are wrong.

PR #130, opened 2015-12-03
Adds a CloudFlare and DNSimple category

aeden:
Also adds a DS article and cleans up other categories a bit.


weppos:
Even if technically the article is about setting up the DS record, I suggest we should rename to something like

> Using CloudFlare DNSEC with DNSimple

It seems more clear to me the intent why you want to add the DS record.

PR #133, opened 2015-12-03
Article about what to do with DS records when transitioning DNS

aeden:
This is potentially a start to an article explaining how to handle DS records when changing DNS providers. Please let me know what else you think should go into this article.


aeden:
Looking for feedback from @aetrion/dnsimple - what questions are left open for you when you read this?

sbastn:
:clap: The only thing missing for me would be a screenshot of where to remove/re-add the DS record.

weppos:
We should probably add a link to this article from:

- The article about troubleshooting the domain not resolution
- The article about transferring a domain without downtime, where we suggest to switch the DNS

PR #136, opened 2015-12-03
Update DNSSEC article.

aeden:


sbastn:
:100: Maybe the only thing missing is a small blurb about what to do with DS records when transitioning DNS, aka https://github.com/aetrion/dnsimple-support/pull/133

weppos:
:shipit: :+1:

aeden:
I can't figure out how to word a part about what to do with DS records when transitioning @sbastn

PR #137, opened 2015-12-03
Clarification

therobot:
A required paid account is necessary to build ssl certs.

This is a small clarification that may need a bit of English rewording but not too much.


sbastn:
:boom: yep! Looks good to me.

weppos:
Thanks for the patch @therobot. I made a small change to avoid using path with the hostname or specific protocol as they belong to the same domain (see 312e9a097dc6a8c967b58e85581d5711209d9b75)

weppos:
I already applied the commit directly on master.

PR #141, opened 2015-12-03
Heroku different app error page

### LINKED ISSUE #135: Article about troubleshooting different app on Heroku errors
Sometimes customers say that they have pointed their domain to their Heroku app and that they get an app that isn't theirs (or is theirs but is not the one they wanted). 

This article would be similar to:

- http://support.dnsimple.com/articles/heroku-error-ssl/
- http://support.dnsimple.com/articles/heroku-error-nosuchapp/

jacegu:
Closes #135 

This one also includes a link to a section that is not in master yet.

### DIFF
--- a/content/articles/heroku-error-differentapp.markdown
+++ b/content/articles/heroku-error-differentapp.markdown
@@ -15,8 +15,17 @@ This error occurs when on DNSimple you point a domain to an Heroku application, 
 The error means that your DNS are correctly resolving to Heroku. Follow the instructions below to fix the error.
 </callout>
 
+This error occurs when on DNSimple you point a domain to an Heroku application, but on Heroku an application is already set up to use that custom domain.
 
-There are two possible scenarios:
+Here is an example that should help understanding what's going on:
+
+You have two Heroku applications: _green-turtle_ and _red-turtle_ with their corresponding Heroku hostnames `green-turtle.herokuapp.com` and `red-turtle.herokuapp.com`.
+
+You add the domain `awesome-turtles.com` in DNSimple and you point the CNAME `www.awesome-turtles.com` to `green-turtle.herokuapp.com`. However, in Heroku you associate the `www.awesome-turtles.com` custom domain to the application `red-turtle.herokuapp.com`.
+
+What will happen is that when someone visits `www.awesome-turtles.com`, Heroku will route the request to the _red-turtle_ app, and not _green-turtle_ as you would expect given your DNS settings.
+
+You can experience the same behaviour if the _red-turtle_ application belongs to someone else which leaves us with two possible scenarios:
 
 #### The application with the custom domain setup is yours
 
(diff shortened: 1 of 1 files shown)


content/articles/heroku-error-differentapp.markdown
```diff
@@ -0,0 +1,29 @@
...
+excerpt: You pointed your domain to Heroku, but when you access your domain you see a different application than the one you expected.
+categories:
+- Heroku and DNSimple
+---
+
+# Troubleshooting Heroku shows different app error
+
+You pointed your domain to Heroku, but when you access your domain from the browser you see a different application than the one you expected.
+
+This error occurs when you point a custom domain to Heroku, but there is another app at Heroku with that domain name set up as a custom domain.
```
weppos: This sentence is a little bit unclear. I'd rephrase it to

> This error occurs when in DNSimple you point the DNS for a domain to Heroku, but in Heroku you associated the custom domains to a different application.

I'd also add an example

> For example, you have two Heroku applications: `green-turtle` and `red-turtle` and the corresponding Heroku hostnames `green-turtle.herokuapp.com` and `red-turtle.herokuapp.com`. You add the domain `awesome-turtles.com` in DNSimple and you point the CNAME `www.awesome-turtles.com` to `green-turtle.herokuapp.com`. However, in Heroku you associate the `www.awesome-turtles.com` custom domain to the application `red-turtle.herokuapp.com`.
> What will happen is that when someone will go to `www.awesome-turtles.com`, Heroku will route the request to the `red-turtle` app, and not `green-turtle` as you would expect.

weppos:
I generally use `curl` to debug the issue. I send an HTTP request to the expected endpoint and I append a custom `Host` header.

```
curl -I -H 'Host: www.awesome-turtle.com' http://green-turtle.herokuapp.com
```

Then I either perform the same request via curl again without the custom header, or I visit the site and I compare if they match.

```
curl -I http://green-turtle.herokuapp.com
```

If they don't, it means there is a routing issue. Does it make sense to add the first curl example as a possible debug hint?

content/articles/heroku-error-differentapp.markdown
jacegu: Adding this as I find it very useful. 

Is there some syntax that I can use to make it clear that it's an example and not a text explanation at the same level of everything else?

jacegu:
I'd like the example to differentiate from the rest of the text. Not sure what syntax to use there. Ideas?

weppos:
We don't have a specific syntax for examples at the moment.

jacegu:
@weppos I'm not sure whether it's worth it to include the curls here. If they end up in this page they already know the displayed app is not what they expected which is what the curl will help them figuring out.

What do you think?

weppos:
:+1:

jcaudle:
I put some phrasing tweaks in, but this looks pretty good. I had no idea this could actually happen. :scream_cat:

jacegu:
:ship: this ?

weppos:
giphy

PR #142, opened 2015-12-03
Setting up accounts for clients

meeunier:
Tips for agencies on how to create and manage multiple accounts for clients

### DIFF
--- a/content/articles/setting-up-accounts-for-clients.markdown
+++ b/content/articles/setting-up-accounts-for-clients.markdown
@@ -26,7 +26,7 @@ The member will also have full-admin access to the account, including the abilit
 
 ## Creating a separate account for your client
 
-To add another account to your user profile, you should take the following steps.
+First, you must add another account to your user profile by taking the following steps.
 
 <div class="section-steps" markdown="1">
 1. Visit the settings for your user profile by clicking the <label>Settings</label> link in the dropdown menu.
@@ -49,9 +49,9 @@ To add another account to your user profile, you should take the following steps
   
 ## Inviting your client to DNSimple
   
-  Before adding your client to the new account, she will need to be a __registered__ DNSimple user. 
+  Verify with your client that is a __registered__ DNSimple.
   
-  If needed, you can [point your customers to the signup page](https://dnsimple.com/signup).
+  If needed, [point your client to the signup page](https://dnsimple.com/signup).
   
   <note>It is not necessary that your client have an active subscription, just that they have already signed up for the service.</note>
   
--- a/content/articles/setting-up-accounts-for-clients.markdown
+++ b/content/articles/setting-up-accounts-for-clients.markdown
@@ -55,7 +55,7 @@ First, you must add another account to your user profile by taking the following
   
   <note>It is not necessary that your client have an active subscription, just that they have already signed up for the service.</note>
   
-## Adding members to an account
+## Adding clients to the new account
 
 <div class="section-steps" markdown="1">
 1. Once you've subscribed the new account to an eligible plan, visit the account page by clicking the name of the account in the dropdown menu.
--- a/content/articles/setting-up-accounts-for-clients.markdown
+++ b/content/articles/setting-up-accounts-for-clients.markdown
@@ -1,6 +1,86 @@
 ---
-title: Setting up accounts for clients
+title: Creating accounts for clients
 excerpt: How to create and manage multiple accounts for clients
 categories:
 - Account
 ---
+
+# Creating accounts for clients
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+<note>
+In order to have multiple members on a single account, you must first [be subscribed to an eligible plan](https://dnsimple.com/pricing) for that account.
+</note>
+
+Creating a separate account is a flexible and secure way to share domain management with your clients.
+
+Keep in mind that when you add a user to an account, the new member will have full access rights on every resource attached to the account, including domains, contacts, SSL certificates, etc.
+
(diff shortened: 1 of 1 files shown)


meeunier:
Folks @aetrion/dnsimple , I need your :eyes: 

eyes

weppos:
@meeunier It may be worth to mention with a `warning` that each client should have its own account, and you can't add two clients to the same account or each client will be able to manage other client's domains.

I also wonder if this is post is better suitable for a blog post or a support article.

meeunier:
@weppos :
> I also wonder if this is post is better suitable for a blog post or a support article.

I agree that the content is based on other support articles but the purpose of this is to **support** customer with clients getting set up.

I am scratching my own itch here as I will now be able to briefly explain the concept and then point customers to the article. 

I personally think it belongs in the support doc.

> It may be worth to mention with a warning that each client should have its own account, and you can't add two clients to the same account or each client will be able to manage other client's domains.

I'm confused. Can you explain further?

weppos:
@meeunier I added a warning example. Please check if the message makes sense and feel free to update it.

> I personally think it belongs in the support doc.

The reason I pointed it out is because I believe it's a very specific article. Any other feedback from @aetrion/dnsimple?

content/articles/setting-up-accounts-for-clients.markdown
```diff
@@ -0,0 +1,90 @@
...
+The member will also have full-admin access to the account, including the ability to access and change billing information and account details. The user will retain his login credentials and 2-factor authentication configuration.
+
+<warning>
+Each customer must have a separate account, otherwise if you add more than one customer to the same account each customer will be able to manage other customers' domains.
+</warning>
+
+
+## Creating a separate account for your client
+
+First, you must add another account to your user profile by taking the following steps.
```
jcaudle: We should probably link to http://support.dnsimple.com/articles/account-multi/#creating-a-separate-account here.

content/articles/setting-up-accounts-for-clients.markdown
```diff
@@ -0,0 +1,90 @@
...
+
+1. Once you've submitted the new account form, the account is created and you are free to activate it by clicking the <label>Select a plan</label>. You may also configure other parts of your account from the top navigation menu before activating your account.
+
+    ![Add New Account details](/files/add-new-account-details.jpg)
+
+</div>
+  
+## Inviting your client to DNSimple
+  
+  Verify with your client that is a __registered__ DNSimple.
```
jcaudle: "Verify that your client has registered with DNSimple."?

content/articles/setting-up-accounts-for-clients.markdown
```diff
@@ -0,0 +1,90 @@
...
+  
+  If needed, [point your client to the signup page](https://dnsimple.com/signup).
+  
+  <note>It is not necessary that your client have an active subscription, just that they have already signed up for the service.</note>
+  
+## Adding clients to the new account
+
+<div class="section-steps" markdown="1">
+1. Once you've subscribed the new account to an eligible plan, visit the account page by clicking the name of the account in the dropdown menu.
```
jcaudle: You should probably replace these steps too with a link to http://support.dnsimple.com/articles/account-users/#adding-members-to-an-account

jcaudle:
I like this article a lot. It might be worthwhile to put a similar post together on why someone might want to do this and the benefits of doing so on the blog, but I think it's good to keep the article of specifically what to do on the support site.

content/articles/setting-up-accounts-for-clients.markdown
meeunier: It would indeed make it easier for maintenance... but, IMO, it would make it rather annoying for anybody following the steps to jump from page to page.

content/articles/setting-up-accounts-for-clients.markdown
meeunier: oopsie.

:troll: ... or I meant that you must be a verified entity called DNSimple to use the feature.

content/articles/setting-up-accounts-for-clients.markdown
weppos: Good point @jacegu. We generally prefer to reduce the duplication as otherwise we'll have to maintain it and it's a cost. :wink:

jcaudle: @meeunier this is the standard we have throughout the support site right now. It's not too big of a burden, especially since many of these tasks don't necessarily need to be given step-by-step if the customer is already familiar with them.

meeunier: I think that:

Value from customer creating accounts for clients > Cost of maintaining the tutorial

Therefore we should make the tutorial for the customers (ie.: as easy as possible to follow). In the end, all the customer care is to get the information needed without efforts... clicking bunch of link = extra efforts.

Thoughts @aetrion/dnsimple ?

meeunier: > these tasks don't necessarily need to be given step-by-step if the customer is already familiar with them

Unfortunately, I don't think it's the case right now. 

Here's customer feedback on setting up an account with multiple users (from https://github.com/aetrion/dnsimple-business/wiki/Customer-Feedback):

>"It was a bit confusing that I had created a "personal" account by signing up, and then had to create a new account for our Organization based on my "personal" account. It was weird that my coworker had to go sign up for DNSimple "personally" before I was able to add him to the organization. Ideally I would've had the option of creating the organization right off the bat, and invite users to the organization. Because now Emmet and I have these useless "phantom" accounts. Which isn't a big deal, it's just one more thing we have to remember / make a mental note of when dealing with DNSimple." - https://dnsimple.com/admin/accounts/52027

Issue #138, opened 2015-12-03
Optimize images automagically when deploying the site


sbastn:
Instead of relying on manual optimization, implement an automatic process. Maybe with https://github.com/toy/image_optim?

sbastn:
Just for fun, here what a first round of optimization did:
```
optimizing: 100.0% (elapsed: 9.9m)
49.67%  48.7K  content/files/2fa-barcode.png
34.86%   9.3K  content/files/2fa-code-expiration.png
47.17%  43.7K  content/files/2fa-login.png
56.89%  35.9K  content/files/2fa-recovery-code.png
42.40%  38.6K  content/files/2fa-recovery-link.png
47.80%  50.4K  content/files/2fa-recovery.png
20.64%  20.6K  content/files/access-user-settings.jpg
44.08%  27.5K  content/files/account-billing-change-payment-link.png
55.33%  62.9K  content/files/account-billing-retry-button.png
 3.97%   3.5K  content/files/account-billing-settings-1.png
 2.99%   2.6K  content/files/account-billing-settings-2.png
 9.48%   4.4K  content/files/account-billing-settings-invoice-1.png
------         content/files/account-billing-settings-invoice-2.png
57.95%  91.9K  content/files/account-billing-view-payment-history-link.png
------         content/files/account-change-email-2.jpg
------         content/files/account-change-email-3.jpg
44.54%  15.9K  content/files/account-domains-with-dns.png
20.55%  20.6K  content/files/account-switcher-1.jpg
20.87%  21.3K  content/files/account-switcher-2.jpg
------         content/files/account-unsubscribe.jpg
------         content/files/add-a-record.jpg
22.70%  81.1K  content/files/add-member-account-link.jpg
23.06%  84.8K  content/files/add-member-added-members-screen.jpg
21.72%  64.4K  content/files/add-member-link-members-screen.jpg
21.66%  64.6K  content/files/add-member-screen.jpg
24.29%  79.4K  content/files/add-new-account-details.jpg
24.47%  73.7K  content/files/add-new-account-form.jpg
21.07%  91.0K  content/files/add-new-account-link.jpg
12.83%   6.5K  content/files/adding-a-domain.png
------         content/files/admin-tab.jpg
------         content/files/autorenew-1.png
------         content/files/autorenew-2.png
------         content/files/cancel-subscription-1.jpg
------         content/files/cancel-subscription-2.jpg
58.87%  24.5K  content/files/certificates-renew-action.png
[...]

PR #143, opened 2015-12-03
Image optimization

sbastn:
Based on #138, this PR adds a rake task to optimize all images before we publish the site. It takes 15 seconds or so to complete the task on my machine. I was able to get it down from 10 minutes by removing a few compression workers that we don't need.


jcaudle:
I wonder if this wouldn't be better as something we check with Travis and fail the build if the images could be optimized? I don't know how feasible that is with the tooling at hand, but It could help keep the repo slimmer if we fixup (via a rebase) commits with large images in them as well.

jcaudle:
> fixup (via a rebase)

Obviously, this would only be something worth doing going forward. I don't think we really want to do this on all of our history...

sbastn:
@jcaudle I'll merge this PR and we can hook this with Travis if we want to. If a slimmer repository is worth the effort, I can run it manually against the repo right now until we figure out a better solution.

weppos:
@jcaudle altering the history of a published repo is never a good idea. Moreover, this repo is public, therefore you may end up causing a lot of trouble to people.

weppos:
@sbastn attaching the task to the deploy means there may be side effects to the repo status (which means changes images) at the time of the deploy.

I suggest to keep the the task in rake and have all of us to be responsible of running it if we add images. In this way the image can be optimized *before* being committed. Otherwise, that will happen is that if 3 of us download the repo and deploy it, what will be deployed is not what we have in the repo.

sbastn:
@weppos there are no change to the repository status because I am optimizing the `output/files` folder which is ignored.

I am OK with making this a manual task against the `content/files`. That's where I started originally.

weppos:
@sbastn please also add a note to the README so that we remember to run the task. Actually... I suggest you to create a `CONTRIBUTING.markdown` file with the information about running the task. GitHub will prompt a reminder each time someone opens a PR.

jcaudle:
> altering the history of a published repo is never a good idea

I agree. That's why I said it'd only be good to do the fixups on incoming commits. The point is moot however given the following clarification:

> there are no change to the repository status because I am optimizing the output/files folder which is ignored

I actually really like this method of optimizing. This allows us to have the higher quality images stored on the repo, but the optimized ones on the site. It also makes the Travis check I suggested, as well as a note on the readme, unnecessary. I say we :shipit:.

sbastn:
I agree with your @jcaudle. 

Optimizing only the `output/files` before a deploy makes more sense for our setup. I'll revert back and ship this.

PR #148, opened 2015-12-10
fix broken URL

meeunier:
Just getting myself a nice little 1-commit lead towards the next OMG Thursday :trollface:


weppos:
Merged, thank!

PR #152, opened 2015-12-31
Optimize only images that need it

sbastn:
This PR is based on the conversation on https://github.com/aetrion/dnsimple-support/commit/7b012e29ddb5964a647f1f5c0024da77bce1ef9e. The `imgoptim` rake task only optimizes images that need it.

Example with 2 consecutive runs:
```
➜  dnsimple-support git:(master) ✗ rake imgoptim
Searching for images to optimize...
Optimized output/files/autorenew-1.png
Optimized output/files/autorenew-2.png
All images optimized!

# Second run
➜  dnsimple-support git:(master) ✗ rake imgoptim
Searching for images to optimize...
All images optimized!
```

It takes a few seconds (~12s on my machine) to search for images to optimize, it should be faster than optimizing everything like we did before.

@weppos test it out and see if you like it. If we decide to run this against the `content` folder, it would work as well.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -12,7 +12,7 @@ task :compile do
 end
 
 desc "Publish to S3"
-task :publish => [:compile, :search] do
+task :publish => [:compile, :search, :imgoptim] do
   puts "Publishing to S3"
   puts `s3_website push`
   puts "Published"
--- a/Rakefile
+++ b/Rakefile
@@ -82,7 +82,11 @@ task :imgoptim do
                                jpegtran: false,
                                svgo: false)
 
-  image_optim.optimize_images!(Dir['output/files/*.*']) do |file|
-    puts "Compressing #{file}"
+  puts "Searching for images to optimize..."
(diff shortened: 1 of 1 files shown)


sbastn:
Ready for your :eyes:, this PR is now 28 days old!!

jacegu:
@sbastn one could say it's _vintaaaaggggge_

weppos:
@sbastn the disadvantages of running the optimization against the `output` folder are the following to me:

- the `output` folder is a cache, and it can be deleted at any time (if a CI env, any time). It means every time the deploy will have to reoptimize all the images
- because of the previous point, the deploy time is still considerably long

    ```
    Published
    rake publish  117.40s user 5.66s system 257% cpu 47.782 total
    ```

- it doesn't prevent the size of the repo to grow because of full-size images.

That said, I'm fine to go ahead and merge this PR with the condition that we may need to review it if it will prove to still be slow, and specifically if we will ever decide to use Travis as a deploy mechanism.

:shipit:

PR #160, opened 2016-01-28
Add section on updating registrant for UK domains

### LINKED ISSUE #149: Nominet domains must have Registrar field updated directly with them
After discussion with eNom, it is now clear that some registrar fields on a Nominet domain (specifically the Registrar name) must be updated through Nominet's registrant portal. We should update the .uk documentation to reflect this.

jcaudle:
This change closes #149

### DIFF
--- a/content/articles/domains-uk.markdown
+++ b/content/articles/domains-uk.markdown
@@ -67,7 +67,7 @@ To transfer a UK domain out of DNSimple, [contact support](https://dnsimple.com/
 
 ## Updating the registrant name
 
-It is possible to change most values in the Whois record for a UK domain, but the registrant name must be changed through Nominet's Online Services (Nominet is the Registry for UK domains). It is possible to [register for Nominet Online Services](https://secure.nominet.org.uk/auth/login.html) with the email address you have in your registrant information and update the value for the registrant name with them.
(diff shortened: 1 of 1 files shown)


content/articles/domains-uk.markdown
```diff
@@ -65,6 +65,10 @@ To transfer a UK domain (i.e. example.uk or example.co.uk) into DNSimple, follow
 
 To transfer a UK domain out of DNSimple, [contact support](https://dnsimple.com/contact) and provide your domain name and your new registrar's IPS tag.
 
+## Updating the registrant name
+
+It is possible to change most values in the Whois record for a UK domain, but the registrant name must be changed through Nominet's Online Services (Nominet is the Registry for UK domains). It is possible to [register for Nominet Online Services](https://secure.nominet.org.uk/auth/login.html) with the email address you have in your registrant information and update the value for the registrant name with them.
```
weppos: @jcaudle in some cases it's still possible for us to change it (e.g. it is, the first time as far as I remember or after a transfer).

I'd add something like

> but the registrant name in general should be changed

I'd also add to contact us if they have any question providing the UK domain name to check and the new value to use as a registrant.

aeden: @jcaudle did you want to update with these changes? Seems like it's the only thing to fix before :ship: ing it.

jcaudle: @weppos I made a change in 8e21c9d. Let me know if you want it to change, but I'm going to ship this for now.

PR #167, opened 2016-03-24
Add note that we offer only authoritative DNS

jcaudle:
This seems like a true "Frequently Asked Question" and worthy of being added here.


aeden:
:shipit:

PR #168, opened 2016-04-06
Clarify email forwarding limits per plan.

aeden:


weppos:
@aeden was the change implemented?

PR #179, opened 2016-07-14
Updates screenshots of support documents to match new UI

arnamak:
There are a lot of outdated screenshots in our support site, this PR will update most of, if not all, the screenshots to match our new UI.


weppos:
Well done! Can I ask you why some screenshots are .JPG and others are .PNG? Should we use a single format?

sbastn:
I use both formats depending on the use case. Generally speaking, I use JPG for photographs, illustrations with gradients, lots of details. PNG are best for logos, simple drawings, or screenshots. For support docs, PNG is a superior format, they loose less information when scaled up/down.

We could move all of them to PNG, but this can be done in another PR (54 out of 192 images are JPG).

Well done @arnamak 👏 👏 👏

arnamak:
Okay that's pretty much everything. There are a few things left that I'd like to do. I want to thoroughly check that all of the text still makes sense with updated UI elements, at a glance it all seems ok but I should probably proofread it. There are also some things that I think can be removed. 

However, I get the sense that we want these changes out there, so I can handle the other stuff in another PR if that makes more sense.

weppos:
@arnamak I'm checking it now. Next time, I suggest to use a browser size around 1280.

W: 2872px | H: 1678px is very large, and makes some screenshot hard to view once scaled. Moreover, screenshots takes more space.

arnamak:
@weppos some of them were a little larger than they should have been simply because I was trying to capture the entire screen so as to not lose context. If it's a problem, I can go through and re-do some of them.

weppos:
@arnamak it's not a big deal, but some screenshots like `2fa-recovery-link.png` (and the various 2FA screenshots) are small, with a lot of extra space around because of the large screen (we fix the max size of the site and we add padding right/left).

weppos:
`content/files/account-change-email-2.jpg` is incorrect. The new screenshot displays the changing of the billing email, whereas the screenshot is part of the flow to update the user email.

weppos:
`account-unsubscribe.jpg` is not referenced anywhere. Am I wrong?

weppos:
`add-new-account-details.jpg` is a little bit misleading. If you read the context where it is used, it references a newly created account, and the selection of a plan. I wonder if it should be updated.

weppos:
Good work @arnamak! There are a few small updates I posted above. I also pushed some updates to existing articles to match the UI.

weppos:
👍  

Please squash merge.

PR #181, opened 2016-07-26
Merge "Advanced DNS" category into "DNS"

weppos:
It's not clear the boundary between the two. We don't really need two
categories here.

@dnsimple/dnsimple any objection?


jcaudle:
🚢🇮🇹

onlyhavecans:
👍 :shipit:

martinisoft:
:shipit:

PR #182, opened 2016-07-26
Updates transfer out doc

arnamak:
Quick update to the screenshots in the "Requesting a Transfer code" document.


weppos:
@arnamak the domains-domain-link.png references the account domain listing. I suggest to use a screenshot from that section instead of a dashboard search, to be coherent with the instructions where the screenshot is in use.

arnamak:
@weppos Okay. There is no longer a search for the account domain listing though, and the search from the navbar displays the results as seen here.

weppos:
@arnamak that's a good point. we need then to be consistent so either update che screenshot and the text of the articles where it's used, or update the screenshot preserving the same area. I'll leave up to you the decision of the path.

arnamak:
@weppos I elected to update the wording, and use the screenshot from the search.

weppos:
👍  looks good to me. Please if the screenshot is used elsewhere, make sure to update all the corresponding articles.

PR #187, opened 2016-08-03
WordPress article

weppos:


jcaudle:
This looks very good @weppos. Is there a reason you didn't mention the Wordpress one-click service? I think it defaults to using `blog.#{domain_name}`, but that could be edited if the entire site will be hosted with Wordpress.com

weppos:
> Is there a reason you didn't mention the Wordpress one-click service?

I didn't recall one. Feel free to add another "steps" section where you cover the 1-click service.

weppos:
@jcaudle let me know if you are going to add the section for WordPress.

jcaudle:
I think this is fine to ship as it is. Since the flow is going to be quite different and the use case is pretty narrow (the service just adds a record to the `blog` subdomain), I don't know if emphasizing the service is worth it here.

PR #188, opened 2016-08-07
New plans article

weppos:
This article provides some information about new plans. It also represents a link for some banners we add in other articles when we identify a feature only available to newer plans.

/cc @dnsimple/dnsimple please review


content/articles/new-plans.markdown
```diff
@@ -0,0 +1,114 @@
...
+# New plans and Legacy plans
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+On June 2015 [we released one of our largedt upgrade to our platform](https://blog.dnsimple.com/2015/06/multiple-accounts-new-plans-and-new-features/), that introduced several new features such as multiple accounts, teams, and regional records. With the release of the multi-account feature we also changed our plans structure.
```
meeunier: > On June 2015

My 🇫🇷 would say 

> In June 2015

🇺🇸, which one is it?

jcaudle: `In` is more correct for a month. If we had a specific day, `On` would be correct.

A second note on this line is that `largedt` (in the link), should be `largest`.

aeden:
@weppos I've completed my review. If you would like another set of 👀  for grammar, then go for it.

weppos:
👍

PR #189, opened 2016-08-08
Update terms of the Student Pack offer

jcaudle:
This change to the offer is not live yet, but will be soon.

@Barbamama and @meeunier, does this look good to you?


jcaudle:
I've removed the screenshot, since I cannot get an updated version right now.

PR #190, opened 2016-08-09
Split "Help my email is not there" from the rest of the email validation

weppos:
I wanted a "simple", direct article that explains what to do. I'm not sure if it is simple enough, but I also don't know how to make it simpler than this.

/cc @dnsimple/dnsimple please review


jodosha:
@weppos It's simple because it states clearly that customers should use "conventional" email addresses from CA. :shipit:

PR #192, opened 2016-08-10
Makes steps clear

sbastn:
Agreeing with this comment:
screen shot 2016-08-10 at 9 17 36 am

This PR removes the vertical bars and adds an horizontal bar instead like this:
screen shot 2016-08-10 at 9 23 10 am


sbastn:
I decided to not add a background since steps can have images. I tried with a few grays, but backgrounds didn't really work out. Let me know what you think about the screenshot above.

weppos:
It seems helpful. I'd say let's :shipit: and see how it goes. Thanks!

PR #194, opened 2016-08-18
Article about enabling and disabling auto-renewal

aeden:


aeden:
Before publishing this I need to remove the old auto-renewal management process from https://support.dnsimple.com/articles/domain-auto-renewal/

aeden:
@dnsimple/dnsimple unless there objections, I am going to merge this.

PR #197, opened 2016-08-26
Add a warning against using desired login emails when creating accounts

jcaudle:
Right now, there is no indication that there is any importance to the email address used when creating an account, but creating a new account will prevent the creation of login credentials with the same email address. This adds a note to prevent confusion.


weppos:
Perhaps we should try to explain better (either inline or in an article) that when you create an account you don't create a user access, instead you create a "container".

weppos:
@jcaudle I feel we should either follow up or close this PR.

jcaudle:
This will be better as in-app messaging. We were not able to verify this was the reason people were confused about the purpose of the email on new accounts, so I agree to close it.

PR #202, opened 2016-09-09
Add Let's Encrypt docs

weppos:
General information about the new Let's Encrypt integration.

I have a few more articles to add, but I'm fine to also merge this one for now and move forward.

/cc @dnsimple/dnsimple for initial review


content/articles/letsencrypt.markdown
```diff
@@ -0,0 +1,81 @@
...
+</note>
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+[Let's Encrypt](https://letsencrypt.org/) is a new [certificate authority](/articles/what-is-certificate-authority) that joined the scenes in late 2015, and become an official member of the CA/B forum in 2016.
```
jcaudle: Did Let's Encrypt already become an official member of the CA/B forum? If so, this should be "...became an official member..."

weppos: @jcaudle feel free to edit inline

jcaudle: I was planning on it, I just didn't know what the state of Let's Encrypt was on this question. 😉

weppos:
@dnsimple/dnsimple I've finished the first batch of Let's Encrypt docs. It is ready for a full review.

Please feel free to edit inline, if you spot any typo or minor change. If instead the change deserves a discussion, then comment here.

I have one specific concern that I want to share. I feel that with the introduction of a new certificate type it is becoming a little bit confusing the use of certificate types to represent products. For example, until today we referred to single-name vs wildcard, but this is no longer true.

We now also have LE that is a SAN, but also a single-name. I wonder what's the best way to represent this information, and avoid confusion in the docs (you can understand better what I mean by looking at the changes).

Should we introduce "explicit product names"? Should these names include the CA? (e.g. Comodo Single-Name, Comodo Wildcard, Lets Encrypt).

weppos:
I also have 2 remaining articles that I'll publish in a different branch. They provide a tutorial to request a new certificate using the traditional CA, or let's encrypt CA.

content/articles/letsencrypt.markdown
```diff
@@ -0,0 +1,87 @@
...
+
+As suggested by Let's Encrypt, the renewal will happen at any time after 60 days (30 days before expiration). 
+
+<note>
+Although Let's Encrypt certificates can be installed manually, the entire process is designed to be fully automated. Therefore, we encourage you to use our [certificate API](/v2/domains/certificates/) to fetch the certificate and install it programmatically.
+</note>
+
+## Products
+
+Let's Encrypt currently provides only one product. They issue only domain-validated, SAN certificates. Let's Encrypt doesn't support wildcard certificates.
```
arnamak: Would it be worth mentioning that we offer an alternative for those wanting wildcards?

content/articles/what-is-ssl-san.markdown
```diff
@@ -0,0 +1,43 @@
+---
+title: What is the Subject Alternative Name (SAN)?
+excerpt: The Subject Alternative Name (SAN) is an extension to the X.509 specification that allows to specify additional host names for a single SSL certificate.
+categories:
+- SSL Certificates
+---
+
+# What is the SSL Certificate Subject Alternative Name?
```
arnamak: I'd consider swapping this section with the "What is a SAN certificate" section beneath it. We link to this page from other places in the docs, typically a user is going to click on those links with the question "what's a SAN certificate" in mind. I feel like diving straight into the background and the X.509 specification might be a little much to lead off with.

arnamak:
Aside from the small review comments I left, looks good to me @weppos 👍

arnamak:
If I'm understanding your concern, @weppos, it's that 'single-name certificate' can refer to both LE single-name certificates as well as Comodo certificates, correct?

I can see how that would create confusion, but perhaps simply prefacing 'single-name' with the CA we're referring to would clear it up (i.e. instead of saying 'single-name SSL certificate' we say 'Comodo single-name' or 'Let's Encrypt single-name' depending on the context, while 'single-name' can refer to both). 

Perhaps a note or paragraph that states 'Both LE and Comodo issue single-name certificates, when relevant we will mention the CA, other mentions of single-name certificates can refer to the certificate type itself'.

jcaudle:
> 'single-name certificate' can refer to both LE single-name certificates as well as Comodo certificates

I think that's part of the problem @weppos identified. The other is that LE doesn't fit into either "single-name" or "wildcard" since it can have more than one name on it (if you are on the right plan). When I was reading these documents I largely saw them as informative about the different types of certificates available in general without a lot of thought about how they fit in with our product offering.

My perspective is that these articles do a great job of explaining the difference between the Common Name and SAN, while our interface does a good job of making clear that you're getting a very different product when purchasing a Comodo certificate vs. requesting a Let's Encrypt certificate and I think it's fine to have a more theoretical and explanatory approach here with a practical approach of making things simple in the app.

Does that address the concern you raised @weppos or am I still missing something about what you're asking?

weppos:
@arnamak and @jcaudle you both narrowed the point. The problem I'm facing is that essentially I'm having hard time trying to reference specific products in various parts of the docs.

That's why I was considering, specifically in this page
https://github.com/dnsimple/dnsimple-support/blob/1fbe00705de5460bf02cc38334e650984d4c6b32/content/articles/ssl-certificates.markdown
to not include Let's Encrypt into the first section of "single name" and instead list 3 products:

- (Comodo) Single-name
- (Comodo) Wildcard
- Let's Encrypt

I am somehow concerned about mentioning the CA directly because we haven't doing it historically (mostly because we changed it at some point), but given that for Let's Encrypt we can't do anything different, I think it would be confusing to have:

- (Comodo) Single-name
- (Comodo) Wildcard
- Let's Encrypt

Another approach is to use:

- Standard (or Traditional) Single-name
- Standard (or Traditional) Wildcard
- Let's Encrypt

or simply

- Single-name
- Wildcard
- Let's Encrypt

although this final version would be kind of confusing, because LE is effectively also a single name.

What do you suggest?

content/articles/getting-started-ssl-certificates.markdown
```diff
@@ -20,14 +20,14 @@ Getting a new SSL certificate is a multi-step process and involves several parti
 
 ## SSL certificate requirements {#requirements}
 
-Before starting, take a moment to review how you will use the certificate and be sure to have access to the server configuration or a feature to install a custom certificate on your server.
+Before starting, take a moment to review how you will use the certificate and be sure to have access to the server configuration, or a feature to install a custom certificate on your server.
```
Barbamama: Where do I go to understand how to use a certificate?

weppos: It depends.

We expect you to know what a certificate is for, as you came to us with the need of a certificate (hence you probably expect to use it for HTTPS).

The sentence essentially tells you "be sure to understand your needs before starting" and also "we ask to check if you have access to the server configuration".

This latter information depends on your infrastructure.

Barbamama: May be you should say that then...please be sure to understand your needs before starting"

aeden: Is there specific behavior you are trying to promote or stop?

weppos: Yes, purchasing a cert when you can't use it (e.g. GitHub Pages, or other services that don't allow you to install third party certs)

jcaudle:
I think my preference is for this option:

- Standard (or Traditional) Single-name
- Standard (or Traditional) Wildcard
- Let's Encrypt

Of "Standard" and "Traditional" I think I'd opt for "Standard". I think I could also support using a phrase like "Let's Encrypt Multi-name", but I don't know if it'd be useful or not. I agree that the current usage is a little strange, but the thing about Let's Encrypt is that, in our product offering, it is distinct from the traditional way we've done things with respect to certificates. We could introduce it as SAN, but I don't think the term is well known enough and we already use SAN elsewhere. The other nice thing about introducing Let's Encrypt as "Multi-name" is that it displays something that's in higher tier plans right in the name of the product if someone wants to upgrade for it.

arnamak:
In the context of the page you gave as an example @weppos I think the 'standard' naming scheme works best, as @jcaudle mentioned. 

I think the most confusing part is the two "kinds" of Let's Encrypt products that we offer. Single-name or multi-name, all LE certs are SAN, correct? And it's us that limits it to `www` + apex on our personal plan yes? If I'm correct, this means that LE offers one product that we're splitting into two depending on a plan, and re-using the term 'single-name' to make it easier for the customer to understand what they're getting. Maybe we could call it "Let's Encrypt Personal" and "Let's Encrypt Professional", or something to that effect.

weppos:
> all LE certs are SAN, correct? And it's us that limits it to www + apex on our personal plan yes?

That's correct.

> If I'm correct, this means that LE offers one product that we're splitting into two depending on a plan, and re-using the term 'single-name' to make it easier for the customer to understand what they're getting. Maybe we could call it "Let's Encrypt Personal" and "Let's Encrypt Professional", or something to that effect.

What if we just call it "Let's Encrypt" at this point, and we tell what the product features are (based on the plan).

I'll try to make some adjustments to the docs.

content/articles/what-is-ssl-san.markdown
weppos: 👍

content/articles/letsencrypt.markdown
weppos: 👍

weppos:
I made some final changes, and incorporated the various feedback. Can you give a final check?

weppos:
👀

arnamak:
LGTM @weppos

PR #206, opened 2016-09-23
Fix the HTTPs redirection examples table

### LINKED ISSUE #205: Wrong table on HTTPs redirector article
This was reported by a customer:

> https://support.dnsimple.com/articles/redirector-https/
>
> I can't figure out what the difference in configuration is between rows 2 and 3. 
> Do you mean for the target in row 3 to be https://example.com? (i.e., no www.)
>
> thanks, Jason

jacegu:
Closes #205

### DIFF
--- a/content/articles/redirector-https.markdown
+++ b/content/articles/redirector-https.markdown
@@ -17,10 +17,10 @@ In other words, it's not possible to use the URL record to redirect the www to t
 This table shows a list of most common redirect options, and whether they are supported by the redirector:
 
 | Source | Target | Supported |
-|--------+--------+-----------| 
+|--------+--------+-----------|
 | http://example.com      | http://www.example.com  | YES | Redirect a host name to another URL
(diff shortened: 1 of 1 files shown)


jacegu:
@weppos I PR'ed this tiny change because I wanted you to proofread it.

content/articles/redirector-https.markdown
```diff
@@ -17,10 +17,10 @@ In other words, it's not possible to use the URL record to redirect the www to t
 This table shows a list of most common redirect options, and whether they are supported by the redirector:
 
 | Source | Target | Supported |
-|--------+--------+-----------| 
+|--------+--------+-----------|
 | http://example.com      | http://www.example.com  | YES | Redirect a host name to another URL
+| http://example.com      | https://example.com     | YES | Redirect a host name to the same host name, but HTTPS
```
weppos: The change is correct, but it should be a no. Such redirect is not supported by the redirector (it will end up to an infinite loop).

jacegu: True that.

PR #208, opened 2016-10-05
Article on generating an access token

aeden:
Explains how to generate an access token through the web site.


weppos:
I'd add a line where we link to https://developer.dnsimple.com/v2/#authentication for further information about authenticating using the token.

aeden:
@weppos how's that? Good to go?

PR #213, opened 2016-10-24
Rework secondary articles

aeden:
Removed old article and removes the term slave from the new articles.


aeden:
Was reviewed in person by @weppos

PR #214, opened 2016-10-25
Article for using secondary DNS with Dyn

aeden:
This article explains how to set up Dyn as a secondary DNS provider to us. It covers both Dyn Standard and Dyn Managed, as they use different IP addresses.


sbastn:
👍 🚢

Issue #217, opened 2016-10-27
Reorganize SSL certificate articles


weppos:
I'd like to reorganize the SSL certificate articles, also in preparation of the Let's Encrypt launch.

Here's a few key goals I have in my mind:
- get rid of all the various ways we referred to "getting an SSL certificate", including _purchasing_, _requesting_, _issuing_, etc in favor of a single easy naming:
  - Ordering an SSL certificate
  - Renewing an SSL certificate
  - Rekeying (or Reissuing) and SSL certificate
  
  The reason is because each of the previously mentioned names seems misleading to me:
  - requesting: you request a certificate, regardless the order type.
  - issue: same of above
  - purchase: same of above
  
  Specifically, a renewal is a new certificate, hence you still purchase, and issue it.
- Use "obtain a certificate" when we generally refer to the action of getting a certificate (either via renewal, new order, or other ways)
- Split the process of ordering/renewing a certificate depending on the CA.

---

Here's a list of changes I'd like to apply:
- [x] (work is on #219) Dismantle the current [Getting Started](https://support.dnsimple.com/articles/getting-started-ssl-certificates/) in favor of a more generic article (I outlined a suggested stub [here](https://github.com/dnsimple/dnsimple-support/commit/094df0986e9609a9c93b3db010e23c374db83363#diff-714899a9dc31f7c0ad0d62356e22e91b). The goal is to outline the essential tasks, and potentially delegate to more specific articles where appropriate:
  - _Request the certificate_ can delegate to two articles (one per CA)
  - _Validate the certificate request_ can quickly explain the purpose of the validation, and perhaps mention that depending on the CA the validaiton is different (or link to the articles). I think, again, we want to have specific information about the validation process in the specific articles
  - _Install the certificate_ can link to the specific installing article
- [x] Clearly provide 3 articles, one for each task:
  - New certificate (Ordering?)
  - Renewal
  - Reissue
  
  I had a stub [here](https://github.com/dnsimple/dnsimple-support/commit/094df0986e9609a9c93b3db010e23c374db83363#diff-67a6ba1d9cd21126e19efe0270501a14) for the New Certificate (I was calling it Requesting).
  
  The goal of these articles is to clearly explain the step-by-step process to perform the action, with appropriate screenshots. For example, the "New Certificate" for Comodo should contain the steps we have right now spread across the Getting Started article where we show how to Order > Configure > Submit.
  
[...]

PR #227, opened 2016-11-17
Add a warning about reissuing Let's Encrypt certificates

jcaudle:
Right now, we do not support reissuing Let's Encrypt certificates. Instead, this document suggests customers order a new certificate.

This is related to #217


weppos:
👍  I'd also mention that reissuing a certificate will count against the rate limiting https://letsencrypt.org/docs/rate-limits/ to avoid the risk of people abusing the reissue process and then being unable to get a certificate at some point.

PR #228, opened 2016-11-17
Switch filenames & links to using Standard instead of Comodo

jcaudle:
In order to match our practice throughout the rest of our support site,
this moves mis-named files and creates redirects for the old paths,
while also updating links throughout the site to use "Standard" instead
of "Comodo" in the path and file names.


jcaudle:
Merging this will close #217.

PR #231, opened 2016-11-29
Updates article for PTR records

arnamak:
Updates the Reverse DNS article to state that reverse DNS is available for Pro/Business subscribers.


weppos:
@arnamak can you please follow up here? Is it ok to merge?

PR #238, opened 2016-12-20
Update missing, outdated or bad descriptions

weppos:
This PR contains an update to all legacy "repetitive" descriptions, or outdated ones.


content/articles/reverse-dns.markdown
```diff
@@ -1,12 +1,12 @@
...
+excerpt: We support reverse DNS zones and we can enable the feature on request.
 categories:
 - DNS
 ---
 
-# Do you support Reverse DNS (i.e. PTR records)?
+# Reverse DNS zones
 
-DNSimple does support reverse DNS entries however we do not enable this feature by default on customer accounts. We'll be happy to enable it for you, however first please ensure that you will actually be able to delegate the portion of the subnet that you would like us to respond for reverse DNS queries.
+DNSimple does support reverse DNS entries (i.e. PTR records) however we do not enable this feature by default on customer accounts. We'll be happy to enable it for you, however first please ensure that you will actually be able to delegate the portion of the subnet that you would like us to respond for reverse DNS queries.
```
weppos: Yes, it's a good point. I believe we added it in the past somewhere, I thought it was here. Or was it a different feature?

We also have a specific feature banner (see https://support.dnsimple.com/articles/regional-records/), I think we should use it for consistency.

content/articles/why-we-dont-do-webhosting.markdown
```diff
@@ -1,19 +1,18 @@
...
+Traditionally, DNS registrars have offered domains bundled with a hosting solution as part of a basic package. This usually translates into giving a few MegaBytes of disk space exposed by an HTTP server and accessible with a FTP client for the user. The idea behind this was that the user could have their personal HTML static site bundled in with the registrar's offerings.
 
 This can cause confusion to people without deep understanding of these technologies. This article explains why DNSimple does not offer web hosting. Some of the reasons are:
 
- * We do one thing and We do it very well, that is offering the best DNS hosting experience. 
- * Today the web is way more diverse and complex. There is a myriad of technologies available for creating websites. These technologies use several pieces of software to work, database servers, web servers, caching services, load balancers, etc. Websites have heavier hardware requirements than some Megabytes of hard disk. 
+ * We do one thing and We do it very well, that is offering the best DNS hosting experience.
```
arnamak: Is this still true? We kind of do a lot of things now.

content/articles/reverse-dns.markdown
arnamak: Oh I think this is the file I just merged a PR for.

content/articles/why-we-dont-do-webhosting.markdown
arnamak: We can maybe be a bit more specific. Perhaps something like:

We prefer to focus our efforts on offering the best ever domain management platform by enabling our customers to set up and connect their domains to popular services that already exist, rather than trying to create our own top-to-bottom solution.

PR #240, opened 2016-12-20
Basic DynamicDNS article

onlyhavecans:
A very basic DDNS article. Not sure how in depth to go here since most of this is standard functionality of our API but I have had a few questions about this.


onlyhavecans:
Also not sure if this goes under API or DNS section

weppos:
@onlyhavecans I'm fine to merge. 👍

PR #245, opened 2016-12-27
Merge domain auto-renewal articles

weppos:
Merge both articles into a single one, and update it according to the
latest formatting style.


weppos:
Thanks @arnamak

PR #248, opened 2016-12-29
Add WHOIS privacy info for .IO

weppos:
.IO supports registry-level WHOIS privacy. This is currently not evident in our UI (there is a ticket to handle it). In the meanwhile, I added the details in the support .IO article so that the entire @dnsimple/dnsimple team is aware of it and we can provide it as reference.


weppos:
> Could you please create a Customer Support scenario to support our side of this as well if one does not exist.

👍

PR #254, opened 2017-01-25
Clean up 60 day lock with a hard wrap

onlyhavecans:
This should make it look less like trash.
Fix whitespace on transfering away.

If there is a better way of blocking info out on the support page I am all ears


weppos:
Thanks @onlyhavecans

PR #255, opened 2017-01-26
Document that URLs are passed on base url redirects

onlyhavecans:


weppos:
It's still slightly confusing to me this section:

> The following redirect will not work, because you cannot add DNS records to single URLs:

As you said, it seems to be badly formulated. What I'd do, is to add the text that you added before that example section, and make it more evident that you can do 1-1 redirect from host A to host B, having path forwarding.

PR #257, opened 2017-02-02
Premium Domains article

therobot:
This article provides a short explanation a la FAQ for premium domains. I'd like to get some 🇬🇧 and 🖊  review.

Thanks.


content/articles/premium-domains.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+## What is a Premium Domain?
+
+A premium domain is a domain that has been considered of special value by a registrar. Premium domains generally are available in the new gTLDs domain space. They are usually more expensive that regular domains, prices going up to several thousand dollars for one domain in some cases.
+
+## What makes a Premium Domain valuable
+
+Premium domains appeal to specific businesses or groups of people providing them very with useful, specific domain names for their activities. Other times they are just popular keywords or a very well known combination of name and domain. This is better illustrated with an example: 
+
+For a company such us Heineken, owning the domain heineken.beer is a very powerful and concise way to market its activity, which as you may know is making and selling beers.
```
weppos: Rather than going with a brand, I'd use a different example. For example, you can mention that combination of single-names and TLDs that have very close relations together (e.g. `ruby.codes` or `madrid.love`) are generally premium domains as they are considered having globally recognized value due to the wording itself.

Another mention I think is worth is that 2 and 3 letters new gTLD domains are generally considered premium domains regardless.

content/articles/premium-domains.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+## How can I register a Premium Domain with DNSimple?
+
+When going to our register process, if you type a premium domain name, you'll see an alert informing that the domain is premium and the price itself.
```
weppos: I'd mention register _or_ transfer

content/articles/premium-domains.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+![Premium Domain Alert](/content/files/premium-domain-alert.png)
+
+## What about renewals?
+
+Renewing a premium domain can cost the same than a registration, but it is up to the registrar to decide that, therefore it depends on every specific gTLD.
+
+If you have any other questions about premium domains [you can send us an email](support@dnsimple.com).
```
weppos: Better link to https://dnsimple.com/contact

weppos:
Thanks for the article, I think this is great! I left a few comments.

weppos:
@therobot I don't see any commit, did you push them?

content/articles/premium-domains.markdown
```diff
@@ -0,0 +1,37 @@
...
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+## What is a Premium Domain?
+
+A premium domain is a domain that has been considered of special value by a registrar. Premium domains generally are available in the new gTLDs domain space. They are usually more expensive that regular domains, prices going up to several thousand dollars for one domain in some cases.
```
martinisoft: Would suggest: "A premium domain is a domain that is priced differently by a registrar and generally found in the new gTLDs space. They are typically more expensive than a normal domain with prices going into the thousands of dollars in some cases. An example of a gTLD is `.love` or `.ninja`"

content/articles/premium-domains.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+## What makes a Premium Domain valuable
+
+Premium domains appeal to specific businesses or groups of people providing them very with useful, specific domain names for their activities. Other times they are just popular keywords or a very well known combination of name and domain. Registrars also consider every two or three leter new gTLDs as premium, there is value on being very short.
```
martinisoft: Would suggest: "Premium domains appeal to specific businesses or groups of people providing them with useful, specific domain names for their activities. Other times they are just popular keywords or a very well known combination of a name and domain. Some registries and registrars also consider two or three letter domains to be more valuable and therefore will be higher priced."

content/articles/premium-domains.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+All of this is better illustrated with an example: Paris is worldwide known as the city of love, the domain created as a combination of paris and love therefore the domain `paris.love` is a very attractive one for marketeers, business in the city or even the city itself and may be flagged by the registry as premium domain.
```
martinisoft: Would suggest: "Here is an example: Paris is known as the city of love. The domain created as a combination of these could be `paris.love` and would be a very attractive name for marketing companies or businesses. The `.love` registry therefore may impose a higher fee for this rare and beautiful combination and the registrar may optionally add to this price to make it more expensive than a typical `.love` domain registration."

content/articles/premium-domains.markdown
```diff
@@ -0,0 +1,37 @@
...
+
+## How can I register or transfer a Premium Domain with DNSimple?
+
+When going to our register process, if you type a premium domain name, you'll see an alert informing that the domain is premium and the price itself.
```
martinisoft: Would suggest: "When going through our domain registration process, if you try to register what is flagged as a premium pricing domain you will see an alert which will indicate the new price versus the typical price for that given domain. The alert will look like the screenshot below."

weppos:
@martinisoft please make the changes directly in the article, so that we shorten the review as we can look at the diff instead of trying to diff it.

Issue #236, opened 2016-12-20
Article about updating account email


weppos:
We have an article that explains how to update the user email, but not account emails.

onlyhavecans:
Closed by #270

PR #270, opened 2017-03-16
Add article for Account Email Settings

onlyhavecans:
This not only walks through how to change but also explains why you
might not be able to change it directly.


weppos:
I believe this closes #236

PR #283, opened 2017-05-16
Add warning to articles related to GH Student pack

meeunier:
This PR adds warnings to articles related to the GH Student pack to notify that we are no longer participating in the program.


meeunier:
I'm not 100% sure about the copy in the warning, feel free to tweak it!

weppos:
I would probably change from `<warning>` to `<note>`

meeunier:
@weppos switched `warning` to `note` as suggested. 👍

Issue #44, opened 2014-12-23
Resolution status article is still referencing the old domain page


weppos:
http://support.dnsimple.com/articles/resolution-status/

weppos:
The screenshots and the content is outdated again.
https://support.dnsimple.com/articles/resolution-status/

PR #295, opened 2017-09-27
Updates to the resolution article

sbastn:
This PR updates the `/articles/resolution-status/` with the following:
- Updated content to clarify the origin of the data and why does it mean when a domain is resolving with us or not.
- Fresh screenshots.


sbastn:
Closes #44

PR #299, opened 2017-10-03
Article about reactivating a subscription

### LINKED ISSUE #280: Article on reactivating a cancelled subscription
Our reactivation process is no longer the same as the activation process. Activation includes the trial, reactivation does not. To that end, we should have a support article specifically for reactivation.

jacegu:
Closes #280 

Article that explains how to reactivate an existing DNSimple subscription. It clearly points out the limitations when reactivating:
- Absence of trial.
- Only newer plans are available.


jacegu:
> The last link doesn't work.

The dashboard one? 😕

doctoraw:
Don't worry. I see now it is the image below.

It looked like a link:
image

 ✓ First time reviewing an article

jacegu:
No worries :-)
Thanks for the review :heart:

PR #315, opened 2017-11-28
Article about payment methods

### LINKED ISSUE #309: Article about payment methods
Explaining:
- We only accept credit card as a payment method (and why).
- We have a Business feature that allows wire pre-payments.

jacegu:
Closes #309 

I think I've addressed the basic information. Let me know if you think anything is missing.

### DIFF
--- a/content/articles/payment-methods.markdown
+++ b/content/articles/payment-methods.markdown
@@ -1,0 +1,34 @@
+---
+title: Payment methods
+excerpt: Payment methods supported by DNSimple.
+categories:
+- DNSimple
+---
+
+# Payment Methods
+
+DNSimple only supports two payment methods:
+
+## Credit card
+
+<note>
+DNSimple doesn't store credit card information. We use a PCI-compliant billing provider to process and store all payment data.
+</note>
+
+When you subscribe to DNSimple you will be required to provide a credit card. This card will be used to collect the payment for the subscription fees and all the purchases that you make in DNSimple. These purchases include:
+- Domain registrations
+- Domain renewals
+- Domain transfers
+- Comodo single-name SSL certificates
+- Comodo wildcard certificates
+
+
+## Wire transfer
+
+<warning>
+This payment method is only available under the Business and Master Reseller subscription plans.
+</warning>
+
+Our [Business plan](https://dnsimple.com/pricing) and [Master Reseller plan](https://dnsimple.com/reseller) customers can issue a wire transfer to charge their DNSimple balance. Once DNSimple receives the money, the account credit will be adjusted, and all charges will be deducted from it.
+
+When the credit runs out, the credit card on file will be charged.
--- a/content/articles/payment-methods.markdown
+++ b/content/articles/payment-methods.markdown
@@ -9,13 +9,13 @@ categories:
 
 DNSimple supports two payment methods:
 
-## Credit card
+## Credit or debit card
 
 <note>
-DNSimple doesn't store credit card information. We use a PCI-compliant billing provider to process and store all payment data.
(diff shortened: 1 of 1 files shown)


content/articles/payment-methods.markdown
```diff
@@ -0,0 +1,34 @@
...
+---
+
+# Payment Methods
+
+DNSimple only supports two payment methods:
+
+## Credit card
+
+<note>
+DNSimple doesn't store credit card information. We use a PCI-compliant billing provider to process and store all payment data.
```
OleMchls: Sidenote: I recently learned that especially with Chargify we also would need a certification to process the CC and fwd them to Chargify.

As we are switching to Stripe where this is not the case anymore. I do not see any reason to change this.

content/articles/payment-methods.markdown
```diff
@@ -0,0 +1,34 @@
+---
+title: Payment methods
+excerpt: Payment methods supported by DNSimple.
+categories:
+- DNSimple
+---
+
+# Payment Methods
+
+DNSimple only supports two payment methods:
```
OleMchls: ~~only~~

_only_ makes this sounds like if it's a bad thing.

content/articles/payment-methods.markdown
```diff
@@ -0,0 +1,34 @@
...
+
+## Credit card
+
+<note>
+DNSimple doesn't store credit card information. We use a PCI-compliant billing provider to process and store all payment data.
+</note>
+
+When you subscribe to DNSimple you will be required to provide a credit card. This card will be used to collect the payment for the subscription fees and all the purchases that you make in DNSimple. These purchases include:
```
OleMchls: IIRC @aeden mentioned that it's not necessarily a _credit_card but could also be a debit card. If so, we should change this so only _Card_

OleMchls: I'd remove the second _DNSimple_ to reduce duplication.

content/articles/payment-methods.markdown
aeden: ✅

content/articles/payment-methods.markdown
```diff
@@ -7,20 +7,23 @@ categories:
...
-When you subscribe to DNSimple you will be required to provide a credit card. This card will be used to collect the payment for the subscription fees and all the purchases that you make in DNSimple. These purchases include:
+When you subscribe to DNSimple you will be required to provide a credit card. This card will be used to collect the payment for the subscription fees and all the purchases that you make in DNSimple. These purchases may include:
+
 - Domain registrations
 - Domain renewals
 - Domain transfers
+- Whois privacy
 - Comodo single-name SSL certificates
 - Comodo wildcard certificates
+- Email forwarding
```
jacegu: Are you sure we should add Email Forwarding? Technically it's not a purchase because it's not paid immediately. I've always considered part of the subscription fee.

jacegu:
@aeden another thing that I considered mentioning, but wasn't sure we want to add, is the possibility of a card-less account. WDYT?

content/articles/payment-methods.markdown
jacegu: 👍

content/articles/payment-methods.markdown
jacegu: ✅ and ✅

content/articles/payment-methods.markdown
aeden: Fair enough, go ahead and remove it.

jacegu:
Merging as is. We can modify it in the future if we want to say anything about other payment methods.

PR #320, opened 2018-01-14
DNSSEC wording updates

onlyhavecans:
* update the secondary wording
* clarify you cannot disable automatic rotation


onlyhavecans:
While slightly redunstant should we have a "updating your DS records manually" section?

PR #330, opened 2018-03-28
Fix markdown for links

coreyhaines:
Apparently markdown for linking isn't supported in `info` tags.


weppos:
It looks like `info` is incorrect, it should be `callout`.
https://github.com/dnsimple/dnsimple-support/blob/master/lib/preprocess_filter.rb#L4

coreyhaines:
Making change to `callout`

coreyhaines:
Changing to callout worked.

coreyhaines:
*sigh* I'm guessing that the github preview doesn't run this preprocessor, so it doesn't show up as the `div` with the appropriate classes. Putting `callout` instead of `info` and keeping fingers crossed.

weppos:
Yep, those are custom tags. You can run it locally if you want.

coreyhaines:
I'll trust the gods of preprocessors.

PR #334, opened 2018-04-18
Update header img

doctoraw:
Related to new headr PR
https://github.com/dnsimple/dnsimple-app/pull/6859


sbastn:
You have a file called `content/files/Group 32.png` Can you please rename it to something more clear? We usually use something like `account-add.png` for example.

PR #343, opened 2018-06-25
GDPR addendum

sbastn:
This PR clarifies that certain ccTLDs do not apply GDPR


aeden:
I agree with @weppos on this: this article needs specifics about which TLDs now default to protected data and which ones would still require whois privacy.

PR #350, opened 2018-09-02
Add articles about adding specific record types

### LINKED ISSUE #232: Articles about adding specific record types
It sounds silly, but simple articles like:

- How to add an A record.
- How to add a CNAME record.
- How to add an ALIAS record.
- How to add MX records.

Might be helpful.

weppos:
This adds a standard set of pages for the common records: A, AAAA, CAA, CNAME, URL.

For each record:

- A generic page explains what is the record, how it works
- A DNSimple specific page covers how to manage the record

We have other record pages that needs to be cleaned up and adjusted to match this structure, but it was a lot of work and I focused only on the main records.

---

A review is probably easier using the live version:

```
$ bundle exec nanoc live
```

Closes #232


content/articles/manage-a-record.markdown
```diff
@@ -16,7 +16,7 @@ categories:
 
 You can manage [A records](/articles/a-record) in DNSimple using the [DNS record editor](/articles/record-editor).
 
-The instructions in this article assumes you are familiar with the [A record format](/articles/a-record#record-format) and usage.
+The instructions in this article assume you are familiar with the [A record format](/articles/a-record#record-format) and usage.
```
weppos: @itsalyse all the 5 `manage-*-record` articles have the same structure. Should this change be reflected in all the other 4 articles?

itsalyse: It should. I'm going through and making it all uniform.

PR #351, opened 2018-09-02
Better explain cname unique record error

OleMchls:
Originating from recent customer feedback this is an appraoch to better explain the "CNAME must be the only record on a subdomain" error.

@reviewers: Please feel free to edit freely

### DIFF
--- a/content/articles/cname-record.markdown
+++ b/content/articles/cname-record.markdown
@@ -31,4 +31,4 @@ For example, if the domain were `example.com` and you wanted `www.example.com` t
 
 To understand the error, it is important to understand that a CNAME points a whole subdomain to another name in the domain name system. If you have another record on that subdomain you can't add a CNAME, as that CNAME would render the other records useless. Let's look at an example:
 
-Let's there is an MX record on email.example.com and now you try to add a CNAME on that exact subdomain (email.example.com) that would produce that error because there is already that MX record.
+Let's assume there is an MX record on email.example.com and now you try to add a CNAME on that exact subdomain (email.example.com) that would produce that error because there is already that MX record.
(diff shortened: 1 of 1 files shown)


OleMchls:
@aeden please have another look :)

content/articles/cname-record.markdown
```diff
@@ -26,3 +26,11 @@ For example, if you have a server where you keep all of your documents online, i
 To use CNAME records, select CNAME from the *Add Record* drop down in the advanced editor. Then enter the hostname you would like to alias from and the fully-qualified domain name you would like to alias to. You may also enter `@` in the Alias for field to represent the domain itself.
 
 For example, if the domain were `example.com` and you wanted `www.example.com` to point to `example.com` you could put `www` in the name field and `@` in the alias for field.
+
+### Error: "Cannot add a new record where a CNAME exists"
+
+To understand the error, it is important to understand that a CNAME points a whole subdomain to another name in the domain name system. If you have another record on that subdomain you can't add a CNAME, as that CNAME would render the other records useless. Let's look at an example:
+
+Let's assume there is an MX record on email.example.com and now you try to add a CNAME on that exact subdomain (email.example.com). That would produce that error because there is already that MX record.
```
coreyhaines: This feels like it is a restatement of the above paragraph. How about something like


> let's assume there is an MX record on email.example.com, and now you try to add a CNAME on that exact subdomain (email.example.com). If you added the CNAME, it would override the subdomain (email.example.com) and render the MX record useless, leading to a lot of potential confusion when email stops. To counter this potential confusion, the domain name system does not allow other records alongside a CNAME.

PR #359, opened 2018-09-28
UK domains may or may not be extended depending on expiration date

aeden:
Clarifies extension policy for UK domains (including co.uk).


aeden:
Thanks @coreyhaines , I agree with the change.

PR #363, opened 2018-10-09
update_icann_verification

ARabbitt:
We have seen several messages come into support for the ICANN verification from donotreply@name-services.com with the link going to http://www.enom.com/raaverification/verification.aspx. This PR addes this information to our support page.

See ticket https://dnsimple.groovehq.com/groove_client/tickets/94018927 as an example.

### DIFF
--- a/content/articles/icann-domain-validation.markdown
+++ b/content/articles/icann-domain-validation.markdown
@@ -13,13 +13,15 @@ When you make a change to your registrant's email address or name, you will rece
 
 ![ICANN Verification Email](/files/icann-verification-email.png)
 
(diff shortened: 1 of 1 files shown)


ARabbitt:
Basically recent verification emails have links that start with http://www.enom.com/... instead of http://raa.name-services.com...

Our document says the link must be http://raa.name-services.com... 

This change is to include these enom links as being valid.

weppos:
@aeden @ARabbitt what should we change here to get this shipped?

aeden:
> @aeden @ARabbitt what should we change here to get this shipped?

Get a second reviewer.

Issue #318, opened 2017-12-20
Alerts style consistency


doctoraw:
From issue:
https://github.com/dnsimple/dnsimple-app/pull/6138 

Alerts in support need to be updated for consistency with the app.
![image](https://user-images.githubusercontent.com/413249/34203499-a576b890-e57b-11e7-9bc1-a8bcb231f2a7.png)

PR #376, opened 2018-11-07
Cleanup Markers

weppos:
Historically we had 3 types of markers (alerts) defined:

- callout
- info
- note

Unfortunately, the use has not been very clear, not even to me that I originally defined it. 😓 
We also had some occurrences where we had `warning`.

So I decided to invest a bit of time to cleanup all the markers sections. I defined 4 markers, adopting the same conventions from https://confluence.atlassian.com/doc/info-tip-note-and-warning-macros-51872369.html which I think are quite clear.

I also had to make some changes to the existing styles (which btw were also coded by me quite long time ago) to properly reflect the marker meanings. @doctoraw @sbastn feel free to step in.


doctoraw:
Related to issue #318

weppos:
@doctoraw what do you want to do with #318? Is it something do you think we should work on?

doctoraw:
@weppos It is related to the design system project. I haven't almost proposed any changes to the support site because I've been more focused on the app. Now that we have a style in place for alerts on the app, we can reuse them for the support site. It's a quick win.

PR #383, opened 2018-11-12
App alert style in support alerts

doctoraw:
Following the design system issue about creating a style for [alerts in the app](https://github.com/dnsimple/dnsimple-app/pull/6138), I'm applying it to support alerts.

It makes sense to have different [alert categories in Support](https://github.com/dnsimple/dnsimple-support/pull/376) because most of them aren't triggered by user's actions, like in the App.

So I'm focusing on aligning their aesthetic to App alerts.

App alert example
alert-app

Support alert example
info-alert-supp


doctoraw:
I've adjusted the margins and colors to fit better.

To review all alerts in context, you can visit these two pages:
/articles/before-transferring-domain/#adding-the-domain-to-dnsimple
/articles/api-access-token

All icons inside alerts are now gray, except the warning icon, which is highlighted so it catch the eye even more.

Two examples:
image

weppos:
@doctoraw can you show an example of the 4 different types of message? tip, info, note, and warning.

doctoraw:
@weppos I'm aware. That's why I removed the ready-for-review tag. I'll change that after lunch.

doctoraw:
Review alerts with titles (h4) and links. 

Example:
image

PR #389, opened 2018-11-21
Update references mixing user and account

doctoraw:
We used User and Account as synonyms in the past, when we didn't have the multi-account feature.

I'm reviewing its use everywhere we use those terms.

The trigger for this PR was a user writing us though support asking how to delete an account. We didn't specify that information anywhere, so I've added it to the _Deleting User_ page.


doctoraw:
image

As per this @weppos comment on slack, I've rolled back the changes that relate to teams and members, until we figure out https://github.com/dnsimple/dnsimple-business/issues/709

I'd upload the rest of the changes as they clarify a bit our use of account and user in the app.

PR #390, opened 2018-11-22
Ordering wildcard certificate

### LINKED ISSUE #262: Article on how to buy a wildcard certificate
I just received a request from a customer for instructions on how to buy a wildcard certificate. There is an article for ordering a [standard](https://support.dnsimple.com/articles/ordering-standard-certificate/) certificate, but I haven't seen a walkthrough type article of buying an wildcard certificate.

So I can't point the customer to an article guiding him.

OleMchls:
As a last step for https://github.com/dnsimple/dnsimple-business/issues/606 this article provides a guide to buy a wildcard certificate as mentioned in #262.

---
fixes #262


content/articles/ordering-a-wildcard-certificate.markdown
```diff
@@ -0,0 +1,47 @@
...
+# Ordering a Standard SSL Certificate
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+Wildcard certificates allow you to secure all subdomains of a domain with a single certificate. Wildcard certificates can make certificate management easier in some cases. We still recommend non-wildcard certificates for most use cases.
```
weppos: > Wildcard certificates allow you to secure all subdomains of a domain with a single certificate.

This is not technically true. It allows to secure an unlimited number of names that belong to a single subdomain level.

It's picky, but it's a key difference. We should also update the wording on https://support.dnsimple.com/articles/ssl-certificates-types/#ssl-certificates-by-secured-domains

weppos: > We still recommend non-wildcard certificates for most use cases.

Reason?

content/articles/ordering-a-wildcard-certificate.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+Wildcard certificates are no different to standard certificates, they only contain an `*` as the subdomain, so please make sure to check out the article about [Ordering a Standard Certificate](/articles/ordering-standard-certificate/) too.
```
weppos: This sentence is potentially incorrect. It uses "standard certificate" as an alternative to "wildcard certificate". However, we've been using the term standard to refer to Comodo certificates.
See https://support.dnsimple.com/articles/standard-vs-letsencrypt/

What you probably wanted to use is single-name vs wildcard.

In fact, the counterpart of https://support.dnsimple.com/articles/ordering-standard-certificate/ is https://support.dnsimple.com/articles/ordering-lets-encrypt-certificate/

content/articles/ordering-a-wildcard-certificate.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+
+## Let's Encrypt or Standard
+
+Both providers support wildcard certificates Let's Encrypt and Comodo. Let's Encrypt certificates are free but are only valid for 60 days, Comodo certificates are valid for one year or longer but also cost $100 per wildcard certificate.
```
weppos: > Both providers support wildcard certificates Let's Encrypt and Comodo.

This sentence is missing something. Is there a typo?

weppos: > Let's Encrypt certificates are free but are only valid for 60 days

This is incorrect. They are valid for 90 days.

Generally speaking, I think we should avoid to repeat information that are already available in the specific article, and are even subject to become outdated (e.g. the price, or the duration).

You can stay more generic, and explain that we offer wildcard certificates via both Let's Encrypt and Standard certificates. To learn more about the differences then you can refer to the article.

content/articles/ordering-a-wildcard-certificate.markdown
```diff
@@ -0,0 +1,47 @@
...
+
+You can learn more about the differences between standard and Let's Encrypt certificates in the corresponding [article](/articles/standard-vs-letsencrypt/).
+
+<info>
+Let's Encrypt wildcard certificates are only available on [specific plans](https://dnsimple.com/pricing).
+</info>
+
+## Use * for a wildcard certificate
```
weppos: This section is the core of the article, and it's what I think should contain the specific information. The way you request a wildcard is different between Standard and Let's Encrypt.

What I'd like to see is one section per certificate type where you say:

- Here's how to order a wildcard for Let's Encrypt (linking to https://support.dnsimple.com/articles/ordering-lets-encrypt-certificate/)
- Here's how to order a wildcard for Standard (linking to https://support.dnsimple.com/articles/ordering-standard-certificate/)

The instructions here are very generic and confusing. They don't provide clear direction on both cases.

content/articles/ordering-a-wildcard-certificate.markdown
OleMchls: This initial part was lend from https://community.letsencrypt.org/t/acme-v2-and-wildcard-certificate-support-is-live/55579 😬 

>  It allows to secure an unlimited number of names that belong to a single subdomain level.

Will update that.

>  > We still recommend non-wildcard certificates for most use cases.
> 
> Reason?

When a wildcard certificate gets compromised an attacker can use the certificate to secure a possibly compromised domain. I'm happy to remove that.

weppos: That makes sense. However, I think in the context of this support page it may be more confusing than helpful (especially when left broadly open as it is now).

I'd vote for remove it. Or you can change it into:

> Wildcard certificates can make certificate management easier in some cases. However, you should avoid them unless you have a specific need such as multi-tenant applications.

OleMchls: ✅

content/articles/ordering-a-wildcard-certificate.markdown
```diff
@@ -0,0 +1,46 @@
...
+
+* TOC
+{:toc}
+
+---
+
+Wildcard certificates allow you to secure an unlimited number of names belonging to a single subdomain level. Wildcard certificates can make certificate management easier in some cases. However, you should avoid them unless you have a specific need, such as multi-tenant applications.
+
+Wildcard certificates are no different from single name certificates. They contain an `*` as the hostname.
+
```
laetitiaeden: Could you put an example there: 
wildcard
I found this image on a blog and I feel that really help to understand.  So not the same of course but something similar?

OleMchls: screenshot 2018-11-28 at 09 17 00

I added this ASCII chart to explain the concept.

content/articles/ordering-a-wildcard-certificate.markdown
```diff
@@ -0,0 +1,54 @@
...
+Let's Encrypt wildcard certificates are only available on [specific plans](https://dnsimple.com/pricing).
+</info>
+
+## Use * for a wildcard certificate
+
+To include any subdomain in a certificate, you need to use an asterisk `*` as the subdomain.
+
+### Let's Encrypt
+
+Here's how to [order a wildcard for Let's Encrypt](/articles/ordering-lets-encrypt-certificate/).
```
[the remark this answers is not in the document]
weppos: I made the changes for you. Adding a screenshot would likely be even more helpful.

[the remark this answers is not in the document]
OleMchls: screenshot 2018-11-29 at 15 40 14

✅

PR #388, opened 2018-11-20
Let's Encrypt Wildcard Certificates

OleMchls:
Once this feature goes into public beta this PR documents the usage.

PR #391, opened 2018-11-23
Tightened copy & active voice edits

OleMchls:
extracted copy edits from #388 by @itsalyse

### DIFF
--- a/content/articles/caa-record.markdown
+++ b/content/articles/caa-record.markdown
@@ -1,5 +1,5 @@
 ---
-title: What is a CAA record?
+title: What's a CAA record?
 excerpt: A Certification Authority Authorization (CAA) record is used to specify which certificate authorities (CAs) are allowed to issue certificates for a domain.
 categories:
 - DNS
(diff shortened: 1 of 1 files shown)


content/articles/caa-record.markdown
```diff
@@ -14,13 +14,13 @@ categories:
 
 ---
 
-## What is a CAA record?
+## What's a CAA record?
```
weppos: This article is part of a series about specific record types. I think if we make this change here, we should apply it to all the other types e.g. https://github.com/dnsimple/dnsimple-support/blob/master/content/articles/cname-record.markdown

Basically, these articles should likely all have the same structure. Do you agree @itsalyse ?

itsalyse: We should. We're using "don't" and "won't" and "I'm" in a lot of others, so "what's" would also be used. I've been changing them as I go through the support site articles.

weppos:
Closing this one as there have been other edits merged on the same pages.

PR #417, opened 2019-01-31
How do I pronounce DNSimple

onlyhavecans:
We get this asked enough I thought it would be fun to have a support article about it.

### DIFF
--- a/content/articles/how-is-dnsimple-pronounced.markdown
+++ b/content/articles/how-is-dnsimple-pronounced.markdown
@@ -1,0 +1,21 @@
+---
+title: How to pronounce DNSimple
+excerpt: "It's not obviously apparent how to pronounce DNSimple. Here is a guide for you"
+categories:
+- DNSimple
+---
+
+# The Name DNSimple
+
+DNSimple is a portmantou of the abbrevation for Domain Name System (DNS) and the English word Simple, because our goal has always been to try to make DNS more Simple.
+However pronoucing it is not always simple.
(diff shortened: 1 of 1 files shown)


OleMchls:
I always explain that

> Here at DNSimple we love sharing, that's why DNS and Simple share the "S"

onlyhavecans:
> It'd be nice to have a closing paragraph.

I'm dry for ideas on what kind of closing paragraph we would have on a support article. Can someone suggest something?

OleMchls:
Here is my free idea for you:

> If your still unsure how to pronounce it, record yourself saying it out loud and give your award-winning support a try by sending along your recording and we will also help you out with this (and any other Domain Management automation question).

onlyhavecans:
I added @OleMchls idea @aeden. Does this wrap up the support article better?

content/articles/how-is-dnsimple-pronounced.markdown
```diff
@@ -0,0 +1,27 @@
...
+
+But pronoucing it isn't always simple.
+
+Here's a basic guide:
+
+## Pronouncing DNSimple
+
+The easiest way to visualize it is D-N-Simple.
+
+Here at DNSimple we love sharing, that's why DNS and Simple share the "S".
```
itsalyse: >Here at DNSimple we love sharing, that's why DNS and Simple share the "S"

This seems like something that makes more sense internally than externally. I think the initial explanation is enough without it.

PR #426, opened 2019-03-08
Domain management edits

itsalyse:
Edits for consistency & clarity. Made changes for active voice & tightened copy.

### DIFF
--- a/content/articles/transferring-domain-between-accounts.markdown
+++ b/content/articles/transferring-domain-between-accounts.markdown
@@ -1,47 +1,47 @@
 ---
-title: Transferring a domain to another DNSimple account
+title: Transferring a Domain to Another DNSimple Account
 excerpt: How to move a domain from one DNSimple account to another.
 categories:
 - Domains
 - Domain Transfers
 ---
 
-# Transferring a domain to another DNSimple account
+# Transferring a Domain to Another DNSimple Account
 
 * TOC
 {:toc}
 
 ---
 
-You can transfer (_move_) a domain between DNSimple accounts at any time via the DNSimple interface. Billing responsibility will be transferred to the new domain account, once the move is accepted. Moving a domain to another account will also transfer all the associated resources, such as DNS records, SSL certificates and email forwards.
+You can transfer a domain between DNSimple accounts any time via the DNSimple interface. Billing responsibility will be transferred to the new domain account, once the move is accepted. Moving a domain to another account will also transfer all the associated resources, such as DNS records, SSL certificates, and email forwards.
 
 <warning>
 Once the move is accepted, you will no longer be able to manage the domain.
 </warning>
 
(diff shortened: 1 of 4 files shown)


content/articles/transferring-domain-between-accounts.markdown
```diff
@@ -1,48 +1,48 @@
...
 
-# Transferring a domain to another DNSimple account
+# Transferring a Domain to Another DNSimple Account
 
 * TOC
 {:toc}
 
 ---
 
-You can transfer (_move_) a domain between DNSimple accounts at any time via the DNSimple interface. Billing responsibility will be transferred to the new domain account, once the move is accepted. Moving a domain to another account will also transfer all the associated resources, such as DNS records, SSL certificates and email forwards.
```
weppos: This article explains the process of transferring the ownership of a domain to another account within DNSimple. It's not a standard transfer process, and each provider has a different definition for it.

We used _move_ to distinguish it from a _transfer_, and avoid people being confused. For instance, the _move_ doesn't require an auth code, nor the registrar to be involved. It's basically a change of ownership record at DNSimple.

This review completely removed the concept of _move_ making it almost identical to a transfer. I'm afraid this may turn to be more confusing.

What do you think @itsalyse @aeden ?

aeden: I approved the PR because I think it's the right thing to do. I've repeatedly seen customers use the term "transfer" to describe both internal and external transfers, so let's use their language as well. Furthermore, the UI calls it "Transfer ownership" so this aligns the support text with the behavior.

The only question I have is whether or not a search for the term "move" will find this article once this change is implemented. In my opinion "move" does still need to be found when used as a search term in the support search, thus if it fails to be found with that term then the word needs to be added back to the text (or metadata?) somehow.

weppos: We can write an info paragraph explaining in one sentence the difference between the standard process and include the term _move_.

itsalyse: I think that's the clearest solution.

weppos: > I think that's the clearest solution.

@itsalyse can you apply this change? It looks like this PR got stuck in this state, I think it would be great if we can make the change and ship it.

itsalyse: It looks like I added it a while ago and then forgot about it. 

> This article discusses how to move domains internally from one DNSimple account to another. This is different from transferring a domain away. To transfer a domain away from DNSimple, see [this article](https://support.dnsimple.com/articles/transferring-domain-away/). 

Does that work to resolve the discrepancy?

weppos: THanks.

PR #430, opened 2019-03-15
Clarify information about email address used for notifications

doctoraw:
A user assumed we were sending the expiration warnings about certificates to the email used for that certificate validation. 

I've updated this article to clarify the fact that we use the email associated to the account, not the products inside. This will be helpful for our users, and for us to reference when they ask us again.

To review it, 
1. visit: /articles/product-expiring-tomorrow-notification/
2. check the paragraph added is:
> Emails about expiration products will be sent to your account email address. Changes in certificate validation emails or domain contact information won’t affect your expiration notifications emails.

### DIFF
--- a/content/articles/product-expiration-notification.markdown
+++ b/content/articles/product-expiration-notification.markdown
@@ -9,6 +9,8 @@ categories:
 
 The **Product Expiration** email contains the list of products (purchased domains, certificates and whois privacies) expiring within 2 months, grouped by expiration range.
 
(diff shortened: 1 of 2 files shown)


doctoraw:
Thanks, @weppos . I've done both changes.

doctoraw:
@weppos In this case it's only extra information about the same topic, not a note. In fact, we have this same information in another article, and it's not a note:
> The notifications are delivered to the email address associated with your account.

https://support.dnsimple.com/articles/expiring-product-email-notifications/

PR #432, opened 2019-03-21
Remove 2FA enforcement beta notice

OleMchls:


OleMchls:
Shipping this with a single review as it's a very simple change

PR #441, opened 2019-06-05
Bring back "contact change 60 day lock"

OleMchls:
Reverts dnsimple/dnsimple-support#357

This will need some more adjustments, regarding when we lock. This depends on the registrar we are using for a registration.

### DIFF
--- a/content/articles/icann-60-day-lock-registrant-change.markdown
+++ b/content/articles/icann-60-day-lock-registrant-change.markdown
@@ -7,7 +7,7 @@ categories:
 
 # 60-Day Lock After Change of Registrant
 
-Due to [ICANN policies that went into effect December 1st, 2016](https://www.icann.org/resources/pages/ownership-2013-05-03-en), making certain changes the registrant contact will lock the domain for 60 days. During that period, it will not be possible to transfer the domain to another registrar.
+For all domains under ICANN management, all [gTLDs](/articles/what-is-tld/) and potentially some [ccTLDs](/articles/what-is-tld/), a new [ICANN policies that went into effect December 1st, 2016](https://www.icann.org/resources/pages/ownership-2013-05-03-en), making certain changes the registrant contact will lock the domain for 60 days. During that period, it will not be possible to transfer the domain to another registrar.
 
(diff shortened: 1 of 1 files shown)


OleMchls:
This article, yet, lacks the information that this only applies to some TLDs (depending on the registrar and if it's an ICANN TLD). Is that something we should consider adding? @aeden @weppos ?

aeden:
> Is that something we should consider adding?

I don't know what this would look like.

OleMchls:
> This article, yet, lacks the information that this only applies to some TLDs (depending on the registrar and if it's an ICANN TLD). Is that something we should consider adding? @aeden @weppos ?

As discussed in a call with @aeden we will mention gTLDs and be vague about others. Also, the _What's a TLD_ article covers the differentiation.

content/articles/icann-60-day-lock-registrant-change.markdown
```diff
@@ -0,0 +1,21 @@
+---
+title: ICANN 60-Day Lock After Change of Registrant
+excerpt: After changing your registration info you cannot transfer your domain for sixty days.
+categories:
+- Contacts
+---
+
+# 60-Day Lock After Change of Registrant
+
+For all domains under ICANN management, all [gTLDs](/articles/what-is-tld/) and potentially some [ccTLDs](/articles/what-is-tld/), a new [ICANN policies that went into effect December 1st, 2016](https://www.icann.org/resources/pages/ownership-2013-05-03-en), making certain changes the registrant contact will lock the domain for 60 days. During that period, it will not be possible to transfer the domain to another registrar.
```
aeden:
```suggestion
For all domains under registries under contract with ICANN, including all [gTLDs](/articles/what-is-tld/), [newGTLDs](/articles/what-is-tld/), and potentially some [ccTLDs](/articles/what-is-tld/), a new [ICANN policies that went into effect December 1st, 2016](https://www.icann.org/resources/pages/ownership-2013-05-03-en), making certain changes the registrant contact will lock the domain for 60 days. During that period, it will not be possible to transfer the domain to another registrar.
```

sbastn:
Removing myself for review since you have 2 approvals :)

PR #443, opened 2019-06-17
Add NS records for subdomain

san983:
From this groove [ticket](https://dnsimple.groovehq.com/tickets/74196) I worked on last week. I realised that the there were no support entry for adding NS records for a subdomain there.

Just a draft, but I think it can be useful to catch any future customer question/request about this.

Thoughts?

# Test

- Run the app locally
- Go to http://0.0.0.0:3000/articles/add-ns-records-for-subdomain/


weppos:
I'll remove myself as already waiting for @itsalyse on the content side, and @aeden are gave green light on the technical side.

PR #444, opened 2019-06-19
Update 2FA support to include line about it being online to sync

SarahCaminiti1:
Put in line per feedback from customer


weppos:
I agree with @OleMchls . This is quite confusing, and actually incorrect. What we should explain, is that in case of issue you should ensure your time is in sync by activating the device feature to sync the time.

However, I don't think it should be in the section where this note was placed, rather in a specific troubleshooting.

SarahCaminiti1:
@weppos Can you give me a little more information about what you mean by activating the device feature to sync the time? Is there a specific feature in the 2FA app that syncs the time?

weppos:
Most devices have a feature that can be enabled to sync the Date and Time automatically. Turning this feature on will ensure the clock stays in sync, and prevents (solve) most 2Fa issues.
See https://squareup.com/help/us/en/article/5288-update-date-time-on-your-device

SarahCaminiti1:
I have made the changes described to the document and added a separate section at the end titled "Troubleshooting"

weppos:
> I have made the changes described to the document and added a separate section at the end titled "Troubleshooting"

@SarahCaminiti1 I don't see the changes. Did you submit them?

PR #451, opened 2019-07-02
Clear CTAs

### LINKED ISSUE #446: Remove Bootstrap
We currently have 2 versions of Bootstrap in this site. One that we load from a CDN, and another one that we host locally. Looking at the use of this framework, we only use the grid. This can be replaced with Flexbox.

sbastn:
This PR adds Clear CTAs back to the app. Since I was adding new HTML and CSS, I took the plunge and removed Bootstrap in the process.

Summary:
- Removed Bootstrap.
- Added Tachyons.
- Reworked the header so the CTA is clear and easy to see.
- Reworked the footer to include a CTA.

Closes #446

## Preview
Screenshot_2019-07-02 DNSimple Help - DNSimple Help

## 🔍 Test cases

- Run `yarn` to get the latest dependencies.
- Run `rm -rf output` to remove the `output` folder. This is what gets deliver when we deploy this static site.
- Run `yarn live`. This regenerates the `output` folder.
- Browse the site: look at an article, and a category.
- Perform a search.
- Open support.dnsimple.com in another tab and compare the pages.

## :shipit: Deployment Pre/Post tasks

None.

## 🚀 Verification

See :mag: test cases above


OleMchls:
The search icon is off on latest chrome Screenshot 2019-07-02 at 13 39 23

sbastn:
Thanks @OleMchls for the review! I fixed the glitch with the icon. The reason for the footer color is that it matches a support campaign that we are currently running on the app. It's a visual clue and way to tie the 2 things together.

san983:
@sbastn, it looks great! I like the fact of adding the "app" footer to support 👍  

Just two minors things I found, the on-select shadow around that search inbox is rectangular in chrome
Screenshot_02_07_2019__10_40

But works ok in Safari
Screenshot_02_07_2019__10_41

---

Also the search input is a bit off when you stretch the window 
Screenshot_02_07_2019__10_45

sbastn:
@san983 Thanks for your review. I have addressed both issues:

1. I removed the outline outside the input box
2. Fixed the layout when the input is in responsive mode

Could you please double check again? 🙏 Thanks in advance!

PR #468, opened 2019-11-04
Update to Ruby v2.6.5

san983:
Moves to Ruby v2.6.5


san983:
Rollbacked as Travis ran into https://github.com/laurilehmijoki/s3_website/issues/323 when upgrading to Ruby v2.6.5

PR #473, opened 2019-12-13
Updates package.json libraries to their latest version

sbastn:
Updates package.json libraries manually while we get Depfu on this repository.

I was able to browse a few support pages. No major problems found.


sbastn:
Merging now.

PR #499, opened 2020-01-10
Support doc writing guide

itsalyse:
Creating a brief set of writing guidelines for support docs. This will ensure that all docs meet the same formatting/tone standards.


itsalyse:
@weppos Can you take a look at these guidelines and let me know if there are any questions or anything that's still unclear?

content/writing guide.markdown
```diff
@@ -0,0 +1,58 @@
...
+
+**Use gender-neutral language**
+
+* Always opt for they/them/you language over she/him.
+* They and their can be used as singular pronouns.  
+* Don't default to masculine pronouns. 
+
+**Tense**
+
+* Always use present tense when applicable. It's stronger.
```
weppos: An example would be really helpful here.

content/writing guide.markdown
```diff
@@ -0,0 +1,58 @@
...
+* Every sub-header (when possible) should use the present continuous form.
```
weppos: Out of curiosity, I've been reading several handbook and I found that a more direct form is generally used. For instance, what about:

> Use the present continuous form in every sub-header, when possible: 

vs

> Every sub-header (when possible) should use the present continuous form.

content/writing-guidelines.markdown
```diff
@@ -0,0 +1,58 @@
...
+
+**Use gender-neutral language**
+
+* Always opt for they/them/you language over she/him.
+* They and their can be used as singular pronouns.  
+* Don't default to masculine pronouns. 
+
+**Tense**
+
+* Always use present tense when applicable. It's stronger.
```
weppos: An example would be really helpful here.

content/writing-guidelines.markdown
```diff
@@ -0,0 +1,58 @@
...
+* Every sub-header (when possible) should use the present continuous form.
```
weppos: Out of curiosity, I've been reading several handbook and I found that a more direct form is generally used. For instance, what about:

> Use the present continuous form in every sub-header, when possible: 

vs

> Every sub-header (when possible) should use the present continuous form.

itsalyse: Generally yes. But in guidelines, leading with the subject makes it easier to skim and find. If I'm talking about sub-headers, putting it nearer to the beginning of the sentence is more helpful when someone is looking for something in a hurry.

itsalyse:
> One section that I am missing, but I can probably contribute to it, is a section explaining when to use the note/info/tip/warning widgets.

With these, it would have to be a case-by-case. If it's going to seriously harm something, close their account, make a massive change, etc. then you'd want one. We could add a section that lays out that these notes/info/tips should be used to highlight the most important information, or function as a TL;DR in articles with more text.

weppos:
> > One section that I am missing, but I can probably contribute to it, is a section explaining when to use the note/info/tip/warning widgets.
> 
> With these, it would have to be a case-by-case. If it's going to seriously harm something, close their account, make a massive change, etc. then you'd want one. We could add a section that lays out that these notes/info/tips should be used to highlight the most important information, or function as a TL;DR in articles with more text.

@itsalyse I am actually referring to something similar to https://devcenter.heroku.com/articles/writing#styled-notices. Do you think it makes sense?

We currently have 4 markers:

https://github.com/dnsimple/dnsimple-support/blob/e78625c382c88920a8ff37d673cce46360b84ad3/lib/preprocess_filter.rb#L4

itsalyse:
That heroku example was exactly the kind of thing I was talking about - sorry if that wasn't clear.

weppos:
@itsalyse what's the next step here. Should we format it properly as article and ship it?

itsalyse:
I'm going to make a couple more edits today/tomorrow - then we should be good to go for the finishing work.

itsalyse:
@weppos I added a widget section with brief explanations and examples

weppos:
> @weppos I added a widget section with brief explanations and examples

Thanks. I renamed it to markers as this is how we call them in the tag.

weppos:
Remember you'll need the [_frontmatter_](https://nanoc.ws/doc/tutorial/) to turn this page into an article. Example:

```
---
title: Writing guidelines
excerpt: Bla bla bla
categories:
- DNSimple
---
```

Also, the content will have to be moved in the folder `/articles` or it won't be found.

PR #504, opened 2020-01-28
Update redirects to eliminate 404s.

dallasread:
As per: https://docs.google.com/spreadsheets/d/1t8Qv6N-Iof-URt28N4cFOgABs5uHxMQwJMLpE4YeGhA/edit#gid=77343041


_redirects
```diff
@@ -1,30 +1,117 @@
...
-/articles/master-slave-support/index.html                 /articles/secondary-dnsimple/
-/articles/purchasing-ssl-certificates                     /articles/ordering-standard-certificate/
-/articles/purchasing-ssl-certificates/index.html          /articles/ordering-standard-certificate/
-/articles/ordering-comodo-certificate                     /articles/ordering-standard-certificate/
-/articles/ordering-comodo-certificate/index.html          /articles/ordering-standard-certificate/
-/articles/enabling-and-disabling-auto-renewal             /articles/domain-auto-renewal/
-/articles/enabling-and-disabling-auto-renewal/index.html  /articles/domain-auto-renewal/
-/articles/dnssec-support                                  /articles/dnssec/
-/articles/dnssec-support/index.html                       /articles/dnssec/
+/articles)/ssl-certificate-hostname                                       /articles/can-upgrade-single-hostname-ssl-certificate-to-wildcard/
```
onlyhavecans: Is this supposed to have the `)`

dallasread: This is on the spreadsheet as linked above, which I believe was proved by Enrico. This might be a reasonable one to leave out, as it is clearly miswritten. Do you think we should _can_ it?

sbastn: The route is listed in the spreadsheet with and without the parenthesis. I would keep both for now.

PR #518, opened 2020-02-28
JS Support Snippet

dallasread:
Related to https://trello.com/c/M8RGLrEV/49-reduce-requests-for-vital-information-that-the-customer-system-could-provide-in-the-initial-request-5-10

To help reduce customer support requests, this PR proposes a new way to give customers more in-context help that could be used across all sites (eg. support, development, dashboard, contact page). It pulls data directly from the support site, so it will always be up-to-date. The "trigger" to open this thing could be a link, button, or even just have it inline on a page.

Feb-27-2020 23-50-17

The JS snippet can be included on any page/domain with a `script` tag and an optional default search term.

```html
<script
    async
    type="text/javascript" src="./snippet.js"
    dnsimple-support-snippet="Let's Encrypt"
></script>
```

You can try it out here: 
https://rawcdn.githack.com/dnsimple/dnsimple-support/24543d9492051a77d15dc1bccb01e79698ae2fce/snippet/dist/index.html

The snippet is small, but the article `.md` data is large~ish, but I personally don't see it as a deal-breaker as there are a few ways we can solve it:

```
snippet.js - 591 KiB
|- ./index.js - 15.6 KiB
|- ./data.json - 570 KiB
```

## Must-Haves
- [ ] Add "Can't find what you're looking for?" functionality. Potentially a contact form?
- [ ] Add CSS so it can be displayed inline on a page (eg. at dnsimple.com/contact) OR we could fix it and have it open with a button/link.
- [ ] CORS will need to be allowed for `http://support.dnsimple.com/snippet.js` and possibly `http://support.dnsimple.com/search.json`.
- [ ] Tests... this started as an experiment for viability, so it wasn't clear this would be necessary.

## Nice-to-Haves
- [ ] Load article data asynchronously. We can use `http://support.dnsimple.com/search.json` so there is no added maintenance burden. By doing this, we can also de-couple it from this repo (or put it in the app repo).
- [ ] Add analytics so we can track how used it is, and potentially if customers get their questions answered?
- [ ] For something like developer.dnsimple.com, it may be interesting to prefix all searches with "api"?

## :mag: QA

- Visit https://rawcdn.githack.com/dnsimple/dnsimple-support/24543d9492051a77d15dc1bccb01e79698ae2fce/snippet/dist/index.html.
- Search for a few things to test its usefulness.

## :shipit: Pre/Post tasks

- [ ] ...

## :shipit: Verification

- [ ] ...


dallasread:
Closing as it was a research branch that we can pull from later if need be.

As a project, this approach is viable (~15kb) and seems useful. Things learned:

- It doesn't need to belong to this repo.
- We _should_ use /search.json to GET the data.
- We may want to track the usage.
- There are some JS string searching libraries which tout their speed, but I found them super slow in comparison to the code in this branch.
- I don't anticipate problems integrating this with a contact form.

PR #525, opened 2020-03-11
Restores Travis config without the deployment step

sbastn:
Restores Travis build along with slack notifications

Screen Shot 2020-03-11 at 11 59 11 AM

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: ruby
+rvm:
+  - 2.6.5
 
 cache:
   - bundler
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,12 @@
+language: ruby
+
+jdk: openjdk8
+
+cache:
+  - bundler
+
+notifications:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: ruby
+
```
weppos: We need to pin to the same version we set in https://github.com/dnsimple/dnsimple-support/blob/master/.tool-versions#L1 and update the .tool-versions to inform to update this file as we do in other (private) apps.

Example:

```
# Update also: .travis.yml
ruby 2.6.5
```

jacegu: +1

sbastn: ✅ Done

sbastn:
Verified that Travis is using the correct Ruby version https://github.com/dnsimple/dnsimple-support/pull/525/checks?check_run_id=521878258

Screen Shot 2020-03-20 at 1 54 55 PM

Issue #529, opened 2020-03-20
Test for invalid chars is not running on all articles


sbastn:
Running `rake test` doesn't check any support articles. Below are the files that we're currently checking:

```
content/sitemap.xml.erb
content/index.html
content/404.markdown
content/robots.txt.erb
content/search.html
```

We need to recursively iterate over the `content` folder. We currently use `Dir["content/*"].each do |path|` on `_test/content_test.rb`

PR #530, opened 2020-03-20
Fixes check for invalid content

sbastn:
Fixes https://github.com/dnsimple/dnsimple-support/issues/529

This PR recursively look for markdown files under the content directory and check for invalid characters.


weppos:
Can you please also apply the same fix to https://github.com/dnsimple/dnsimple-developer/blob/master/_test/content_test.rb#L19

PR #536, opened 2020-03-27
.XXX Domain Update

### LINKED ISSUE #290: XXX special policies
The XXX TLD requires a _Member Auth Token_ to be provided in order to activate the domain.

You can provide the token during the registration, as extended attribute:

screen shot 2017-08-15 at 11 51 56

For API, you need to pass those information as extended attributes. We pass those information in the extended attribute API for the XXX, but apparently they are optional, hence users may create a domain without passing that information.

```
curl -X GET \
  https://api.dnsimple.com/v2/tlds/xxx/extended_attributes \
  -H 'accept: application/json' \
  -H 'authorization: Bearer TOKEN' \
  -H 'cache-control: no-cache' \
  -H 'content-type: application/json'
```

```json
{
    "data": [
        {
            "name": "xxx_memberid",
            "description": "Member Authorization Token",
            "required": false,
            "options": []
        },
        {
            "name": "xxx_resolving",
            "description": "Website Resolves",
            "required": false,
            "options": [
                {
                    "title": "Yes",
                    "value": "1",
                    "description": "Website resolves"
                },
                {
                    "title": "No",
                    "value": "nonResolver",
                    "description": "Website does not resolve"
                }
            ]
        }
    ]
}
```

We should document these specific XXX requirements in a separate TLD page, as we do for other TLDs. We should also inform the customer that for member ID changes they have to contact us as we don't currently expose this feature via API or UI.

kayla-broussard:
Closes #290 

Adds a support page for the .XXX TLD, including where to receive a Member Auth Token and how to register this TLD with one both through the UI and through the API. 

### Verification

- [ ] https://support.dnsimple.com/articles/domains-xxx/ exists and provides necessary information.


content/articles/domains-xxx.markdown
```diff
@@ -0,0 +1,66 @@
...
+This article explains the requirements and special procedures for .XXX domain names.
+
+
+## Registration
+In order for a .XXX domain to be activated by the .XXX registry and be allowed to resolve, you must register on [the ICM Registry](http://icmregistry.com/about/sponsored-community/) and receive the required .XXX Member Authorization Token to activate your domain. The .XXX TLD requires a Member Auth Token to be provided in order to activate the domain.
+
+You can provide the token during the registration as an extended attribute:
+![](/files/xxx-member-auth-token.png)
+
+### Registration via the API
```
weppos: I am a bit confused about this section. Is this expected to show an example of an XXX registration via API? If that's the case, that request is wrong as that is the request to show the extended attributes for XXX, not register one.

I would probably remove it and instead simply extend this sentence:

> You can provide the token during the registration as an extended attribute:

to something like

> You can provide the token during the registration in the UI or as an extended attribute via API:

We should perhaps integrate a bit of explanation of what `xxx_resolving` is (see the link I provided you).

kayla-broussard: I've pushed a new commit - let me know if you'd like any changes!

PR #564, opened 2020-05-28
Removes outdated file

kayla-broussard:
Removes the outdated heroku-services.jpg file.


weppos:
@kayla-broussard please go ahead, no need to wait for my review as long as you verified the correct loading of the assets and that not required assets are removed.

PR #585, opened 2020-07-21
Adds warning that these two features do not work in conjunction

kayla-broussard:
This adds a warning to the top of the pages for both DNSSEC and Secondary DNS, warning they user that they cannot be used in conjunction with one another.

### DIFF
--- a/content/articles/dnssec-and-secondary-dns.markdown
+++ b/content/articles/dnssec-and-secondary-dns.markdown
@@ -1,0 +1,23 @@
+---
+title: Why DNSSEC and Secondary DNS may not work together
+excerpt: A detailed explanation as to why DNSSEC and Secondary DNS may not be compatible together.
+categories:
+- DNS
+---
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+## Why won't DNSSEC and Secondary DNS work together?
+
+Here is why DNSSEC and secondary DNS may not work together:
+
+All authoritative name server responses should include the RR set signatures (RRSIG) corresponding to any published DNSKEY that is the Zone Signing Key (ZSK) type. This is necessary because a resolve could get the ZSK from one authoritative and the RRSIG from another authoritative. To include the RRSIG for all ZSK DNSKEYs, the authoritative name servers must have all private key material. We currently do not share private key material, and even if we did, AXFR does not support transferring private key material and thus key rotations would have to be done manually in a coordinate fashion.
+
+Note that in practice, it is possible to run multi-provider DNSSEC without sharing ZSK private key material, however it is not guaranteed to work due to the condition described above with resolvers getting the DNSKEY and the RRset + RRSIG from different authoritative name servers.
+
+You can read more information about multi-provider DNSSEC in this [Draft RFC here](https://tools.ietf.org/html/draft-ietf-dnsop-multi-provider-dnssec-05).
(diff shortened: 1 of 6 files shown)


weppos:
I am swapping myself with @aeden for review, as he can better judge if the level of information we provide in the warning is sufficient. I think we can be slightly more specific (and technical) about "as they will not work in conjunction", but I'll let him determine it.

content/articles/secondary-dns.markdown
```diff
@@ -7,6 +7,10 @@ categories:
 
 # Secondary DNS
 
+<warning>
+  You cannot set up Secondary DNS if you have [DNSSEC](/articles/dnssec) enabled, as they will not work in conjunction. Please ensure that you are not currently using DNSSEC, or have this feature disabled before using Secondary DNS.
```
aeden, line 11:
```suggestion
  You cannot set up Secondary DNS if you have [DNSSEC](/articles/dnssec) enabled, as they will not work in conjunction. Please ensure that you are not currently using DNSSEC, or disable DNSSEC before using Secondary DNS.
```

content/articles/dnssec.markdown
```diff
@@ -15,6 +15,10 @@ This article describes a feature in Public Beta.
 This article describes a feature only available on [newer plans](/articles/new-plans#newer-plans-only).
 </info>
 
+<warning>
+  You cannot enable DNSSEC if you have set up [Secondary DNS enabled](/articles/secondary-dns), as they will not work in conjunction. Please ensure that you are not currently using Secondary DNS, or have this feature disabled before using DNSSEC.
```
aeden, line 19:
```suggestion
  You cannot enable DNSSEC if you have set up [Secondary DNS enabled](/articles/secondary-dns), as they will not work in conjunction. Please ensure that you are not currently using Secondary DNS, or disable Secondary DNS before using DNSSEC.
```

content/articles/secondary-dns.markdown
kayla-broussard: Thank you! And agreed, they should be a separate article that can be linked to from the warning.

content/articles/dnssec-and-secondary-dns.markdown
```diff
@@ -0,0 +1,23 @@
...
+* TOC
+{:toc}
+
+---
+
+## Why won't DNSSEC and Secondary DNS work together?
+
+Here is why DNSSEC and secondary DNS may not work together:
+
+All authoritative name server responses should include the RR set signatures (RRSIG) corresponding to any published DNSKEY that is the Zone Signing Key (ZSK) type. This is necessary because a resolve could get the ZSK from one authoritative and the RRSIG from another authoritative. To include the RRSIG for all ZSK DNSKEYs, the authoritative name servers must have all private key material. We currently do not share private key material, and even if we did, AXFR does not support transferring private key material and thus key rotations would have to be done manually in a coordinate fashion.
```
aeden, line 19:
```suggestion
All authoritative name servers MUST sign all record sets with all private keys that are Zone Signing Key (ZSK) type. This is necessary because a resolver could get the ZSK from one authoritative name server and the RRSIG from another authoritative name server. To include the RRSIG for all ZSK DNSKEYs, the authoritative name servers must have all private key material. We currently do not share private key material, and even if we did, AXFR does not support transferring private key material and thus key rotations would have to be done manually in a coordinated fashion.
```

PR #590, opened 2020-07-29
Update soa-record.markdown

SarahCaminiti1:
Proposed changes to SOA article


content/articles/soa-record.markdown
```diff
@@ -33,3 +35,4 @@ The SOA record includes the following details:
 - The negative result TTL (for example, how long a resolver should consider a negative result for a subdomain to be valid before retrying).
 
 You can't configure these values yourself.
```
SarahCaminiti1, line 37: Can we add something more explicit here?
```suggestion
You can't configure these values yourself. 
```

PR #597, opened 2020-08-03
Update cname-record.markdown

SarahCaminiti1:
CNAME edits


kayla-broussard:
> Line 35 - 40: @kayla-broussard will the Record editor prevent you from pointing to the IP address, or does it only stop you on the second restriction?

Yep, it'll warn you that it can't be an IP address. The sidebar on the right of editing the CNAME also encourages you to go to the support article to learn the difference between A, CNAME, ALIAS and URL records. It also warns you that CNAMES must be unique to other records 👍

SarahCaminiti1:
@itsalyse can we add a comment that the editor will guide you per what @kayla-broussard said above on lines 35-40

PR #598, opened 2020-08-03
Update alias-record.markdown

SarahCaminiti1:
Edits to the document


itsalyse:
> @itsalyse we need to add a little more information about why this was created.
> 
> Something like:
> 
> We created this because Heroku uses dynamic IP addresses . ALIAS records are necessary to allow you to point your root domain to another domain.

I'm not clear on why this is necessary when that's all illustrated in the first section already - especially with addition of the link to the apex domains article. Having both feels like overkill to me. 

This is basically saying the same thing:
"An **ALIAS** record is a virtual record type DNSimple created to provide [CNAME](/articles/cname-record)-like behavior on [apex domains](/articles/domain-apex-heroku/) .

For example, if your domain is example.com, and you want it to point to a host name like `myapp.herokuapp.com`, you can't use a CNAME record, but you can use an ALIAS record."

kayla-broussard:
> Line 30: @kayla-broussard Is this the correct link for "secondary DNS setup process: https://support.dnsimple.com/articles/secondary-dns/

Sorry I missed this @SarahCaminiti1 - yes, that link is correct.

PR #589, opened 2020-07-28
Redirector and push edits

kayla-broussard:
As per dnsimple/dnsimple-business/#1146 

Minor updates to redirector and domain push articles. Larger updates to domain-related articles.

PR #600, opened 2020-08-03
Update redirector.markdown

SarahCaminiti1:
Edits to article


itsalyse:

PR #608, opened 2020-08-03
Update registering-domain.markdown

SarahCaminiti1:
Removed second section of the document per conversation with Ole. Updated walkthrough to match current UI


itsalyse:

PR #611, opened 2020-08-04
Update differences-a-cname-records.markdown

SarahCaminiti1:
Small change


itsalyse:
The plural form "One or more IP addresses" is correct.

PR #629, opened 2020-08-17
Update account-users.markdown

SarahCaminiti1:
Made some edits

- member and user was used interchangeably and it was confusing. Since we call the user a "member" on the app, I changed all of the users to member, accept for one where I said user, or member, to define what we mean.

### DIFF
--- a/content/articles/account-users.markdown
+++ b/content/articles/account-users.markdown
@@ -1,11 +1,11 @@
 ---
-title: Multi-User Accounts
-excerpt: How to manage the users on a DNSimple account.
+title: Multi-Member Accounts
+excerpt: How to manage the members on a DNSimple account.
 categories:
 - Account
 ---
 
-# Managing Multiple Users on One Account
+# Managing Multiple Members on One Account
 
 ### Table of Contents {#toc}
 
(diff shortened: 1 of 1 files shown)


itsalyse:
We call them "users" in a lot more places in support:
 https://support.dnsimple.com/search/?q=user 
and across the app: 
https://github.com/dnsimple/dnsimple-app/search?q=user&type=Code

I'm not sure this is a quick fix, or that we should implement it without more opinions.

SarahCaminiti1:
@itsalyse I prefer "user" to "member" but it's called "member" in the app...so it's a little confusing

itsalyse:
After doing a little more thinking/research, I don't think this is something we want to (or need to) change. There are places where it's more appropriate to use "user" or "member", and places where it sounds flat out weird to say "member". After doing some research, it looks like it's not against best practices to use these terms interchangeably – Google, Microsoft, Stripe, and Zoom all do it, too.

itsalyse:
I edited it so it changes all instances of "user" to "member" when talking about people associated with a multi-user account, but leaves them as "user" if they're not part of it yet/were removed. @kayla-broussard do you have thoughts and/or opinions on this? 

I also think that if we do want to streamline it, we should change instances of "member" to "user" in the app. It makes more sense, and it's significantly easier to do that than to change all instances of "user" to "member".

kayla-broussard:
> I edited it so it changes all instances of "user" to "member" when talking about people associated with a multi-user account, but leaves them as "user" if they're not part of it yet/were removed. @kayla-broussard do you have thoughts and/or opinions on this?
> 
> I also think that if we do want to streamline it, we should change instances of "member" to "user" in the app. It makes more sense, and it's significantly easier to do that than to change all instances of "user" to "member".

I'd agree with all of this 👍 I wouldn't wholesale change all instances within the support repo, and I think this strikes a balance. To address your second point, there would also be times that "member" within the app wouldn't make as much sense as "user".

SarahCaminiti1:
I think this is great! Thank you both!

PR #634, opened 2020-08-20
Add note about how issuewild takes precedence over issue

ggalmazor:
Hopefully, this will prevent customer support requests due to the way that having one issuewild CAA record forces users to create issuewild records for every other cert provider that they would want to get wildcard certs from.


kayla-broussard (approved):
Made some minor line edits (and @itsalyse will likely make more 😂 ) but LGTM!

PR #643, opened 2020-09-02
No more multi-year certificates

weppos:
Remove or update references to multi-year SSL certificates. The maximum allowed validity is now 1 year.

I've search through the docs and I believe I updated all the references. A second eye would definitely be useful to make sure I did not miss anything.

Belongs to https://github.com/dnsimple/dnsimple-business/issues/1139


kayla-broussard (approved):
LGTM 👍 I had a quick search and I don't see any articles missed.

itsalyse (approved):
Approving this copy update as a whole 👍 

The LE article needs more updates for tone/consistency with the other articles, but I'll do a separate PR for that later.

PR #646, opened 2020-09-09
Updates with a warning that wildcard certificates do not cover root

### LINKED ISSUE #620: Increase clarity as to when you need to order both a wildcard and root certificate separately
This was a suggestion from support.

When purchasing a wildcard Let's Encrypt certificate it does not also issue for the root of the domain. This can cause some confusion when someone orders *.example.com and then their URL record for example.com does not work.

There may be more cases where you need to order both a root certificate and a subdomain one.

kayla-broussard:
Closes #620 

Updates the article with a clear warning that a wildcard certificate will not cover the domain apex, and if a user wishes to secure both the apex and subdomains, they will need an additional certificate.

### DIFF
--- a/content/articles/ordering-wildcard-certificate.markdown
+++ b/content/articles/ordering-wildcard-certificate.markdown
@@ -28,7 +28,7 @@ wildcard
 ```
 
 <warning>
-When ordering a wildcard SSL certificate, the wildcard (*) will _only_ secure subdomains. A wildcard certificate cannot secure the root (such as example.com). If you want to secure both the root domain and subdomains, you will also need to [order an additional certificate](articles/getting-started-ssl-certificates/).
+When ordering a wildcard SSL certificate, the wildcard (*) will _only_ secure subdomains. A wildcard certificate cannot secure the root (such as example.com). If you want to secure both the root domain and subdomains, you will need to [order an additional certificate](articles/getting-started-ssl-certificates/).
 </warning>
 
 ## Let's Encrypt or Standard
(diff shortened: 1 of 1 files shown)


itsalyse (approved):

weppos:
@kayla-broussard this is actually a bit more complex, and this change is potentially misleading. 

A certificate is valid ONLY for the names it is attached to. So generally speaking, if you request `example.com` you will get a certificate that covers `example.com`. Likewise, if you request `foo.example.com`, you will get a certificate that covers `foo.example.com` only.

This rule equally applies to wildcard, as discussed here.

HOWEVER, in the ONLY case of Sectigo/Comodo, they actually "expand" the list of names with some default. This is explained in this support page https://support.dnsimple.com/articles/ssl-certificate-names/, section "Root domain". Specifically:

1. If you request www.example.com, they issue a certificate that covers both www.example.com AND example.com
2. If you request *.example.com, they issue a certificate that covers both *.example.com ANd example.com

So the warning @onlyhavecans is referring to is actually just applicable to Let's ENcrypt certificates, where you have full control of the naming and where no request manipulation or adjustment is performed.

Placing the warning where you added it and/or how you phrased it, may be confusing as Standard wildcards don't allow you to add a second name, and will actually cover apex by default.

kayla-broussard:
> Placing the warning where you added it and/or how you phrased it, may be confusing as Standard wildcards don't allow you to add a second name, and will actually cover apex by default.

@weppos Thank you so much for alerting me to this, as I was not aware this was specific only to Let's Encrypt. I will update the article appropriately.

onlyhavecans:
Is this still pending updates?

kayla-broussard:
Apologies @onlyhavecans - this is ready for review 👍

itsalyse (approved):

onlyhavecans (approved):

Issue #636, opened 2020-08-21
Expose search function for 3rd party consumption


dallasread:
While working on https://github.com/dnsimple/dnsimple-app/pull/13857#pullrequestreview-453156768, @OleMchls mentioned that it would be a good idea to show the same results in the support widget AND on the support site. The original intention of the aforementioned PR was to implement specific search results for certain keywords.

@kayla-broussard and [I](https://github.com/dnsimple/dnsimple-app/pull/13857#pullrequestreview-453280700) noticed... instead of `search.json`, we could expose `search.js`, which will contain a simple function that returns the articles _exactly_ how we want them (dictionary-applied and sorted). 

```js
window.DNSimpleSupport.search('keyword', function(err, data) {
  // do something with the data.articles
});
```

Upon completion, both the support site AND the support widget will be showing the same curated results in the same order.

## Tasks

- [x] Expose `window.DNSimpleSupport.search` at support.dnsimple.com https://github.com/dnsimple/dnsimple-support/pull/648
- [x] Add dictionary to `window.DNSimpleSupport.search` https://github.com/dnsimple/dnsimple-support/pull/648
- [ ] Use `window.DNSimpleSupport.search` in the support widget (@kayla-broussard has a local branch with the codez ready to go – need to merge https://github.com/dnsimple/dnsimple-support/pull/648 first)

PR #648, opened 2020-09-11
Add `search.js`

dallasread:
The support widget in `dnsimple-app` is consuming `search.json` from this repo and building out it's own results. This PR moves the article search/filter/sorting logic from the support widget to this repo. The point of doing this is to keep the logic in one place (and add a dictionary of search terms we should swap out for better results). In short, it exposes this (which can now be used by the support widget in `dnsimple-app`:

```js
window.DNSimpleSupport.search('dns')
```

Fixes https://github.com/dnsimple/dnsimple-support/issues/636

## 🔢 To The Reviewers

- I used old-browser-compatible JS syntax (eg. no es6).
- Remove `lunr` as a dependency.
- I moved search javascript to the default layout, allowing for live search from any page.
- Fixed a typo for a .md link of which `nanoc` was complaining.
- We did not have a JS testing thing in place, so I added `jest`.
- We did not have a JS linter thing in place, so I added `eslint`.
- I left `search.json` in tact so that the support widget in `dnsimple-app` keeps working until we update it to use the new search

## :mag: QA

- `yarn && yarn live`
- Visit http://localhost:3000/
- Search for "close", "delete domain", "twofa", etc – you should see reasonably good results along the way
- At some point, press <Enter> when you're in the search box – you should be sent to /search with the proper results
- Try to break it

## :shipit: Post tasks

- [ ] Ensure https://support.dnsimple.com/search.js doesn't do this: `â€™`. It seems to be a weird browser quirk when you visit the JS file directly – substitutes out `’`, but is fine through CURL and works correctly when loaded in HTML.

## :shipit: Verification

- [ ] https://support.dnsimple.com/ should have live searching


content/articles/edns-client-subnet.markdown
```diff
@@ -19,7 +19,7 @@ This article explains what EDNS Client Subnet support is, and how it's configure
 
 ## How ECS Works:
 
-EDNS Client Subnet support is a mechanism defined in [RFC7871] (https://tools.ietf.org/html/rfc7871) for recursive resolvers to send partial client IP address information to authoritative DNS nameservers.
+EDNS Client Subnet support is a mechanism defined in [RFC7871](https://tools.ietf.org/html/rfc7871) for recursive resolvers to send partial client IP address information to authoritative DNS nameservers.
```
dallasread, line 22: `nanoc` was complaining about the space.

.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
dallasread, line 1: Added `eslint` because javascripts are present in this repo

.travis.yml
```diff
@@ -4,8 +4,20 @@ rvm:
...
+  - nvm install $(grep 'node' .tool-versions | cut -d ' ' -f2)
+  - npm i -g yarn
+  - yarn
 
 notifications:
   email: false
   slack:
     secure: YE9LirevstqJ9qglWo2C1IrFEObyThgS72ks0oikEyqKyx1UeEljX3OPX03k+fref/3Z2smtAMH2OgCaXi3cpLIDJI3J14EXOyzMHYz1ul1h52Qp5UrwbKIikD4Yi+RywVLGhLZMedKKI1D0MwcHlwSEuATyIJpZ7do28lEcuL0=
+
+script:
```
dallasread, line 21: Run tests and lint in Travis

content/search/dictionary.json
```diff
@@ -0,0 +1,14 @@
+/*
```
dallasread, line 1: Comments in JSON! 🎉

_headers
```diff
@@ -1,2 +1,4 @@
 /search.json
   Access-Control-Allow-Origin: *
+/search/index.js
```
dallasread, line 3: We want to be able to access this from a different domain that support.dnsimple.com.

san983 (approved):
LGTM. Looked thru the changes and also spent some time playing with the search. Nothing to add/change. ✅

dallasread:
@san983 What about the second review? :trollface:

_headers
OleMchls, line 3: Why not just `/search.js` that'd correspond to `/search.json`?

dallasread: OK, you convinced me. 😄

lib/search.rb
```diff
@@ -17,8 +26,25 @@ def create_search_index
...
+    File.write(index_file, json)
+  end
+
+  def write_js_file(json)
+    js_file = File.join(@config[:output_dir], "search/index.js")
+    js_lines = File.open(js_file).readlines.map(&:chomp)
+    js_lines[1] = "  var ARTICLES = #{json};"
+    js_lines[2] = "  var DICTIONARY = #{dictionary};"
+    File.write(js_file, js_lines.join("\n"))
+  end
```
OleMchls, line 43: This seems very brittle. A change to the .js file may screw up this method. But that dependency is pretty much invisible. Could we achieve the same using ERB?

Something along the lines of 

```ruby
require 'erb'

simple_template = File.read(js_file)

renderer = ERB.new(simple_template)
File.write(js_file,renderer.result())
```

OleMchls: Alternativly passing in `ARTICLES` and `DICTIONARY` via json as well.

OleMchls (commented):
I completed the verification steps on the Netlify preview and the search performs like a charm ✨ 

I think the search-and-replace generation/manipulation of code will be hard to maintain and debug in the future. Maybe we can find a more reliable way to doing this?

lib/search.rb
dallasread: Great idea – I like that mucho better! Thanks!

dallasread: I did not realize there was a Netlify preview link in the branch checks. 🤯

dallasread: So, I had an awful time implementing this because I was developing it by visiting `search.js` in a browser. It took me _too_ long to figure out there is a weird browser encoding quirk that encodes special characters when accessing a `.js` file directly. Everything works fine, just the source code looks funny. Of course, the code works when included in an HTML file though. 🤷

OleMchls (approved):
Thanks @dallasread <3 I think this look way more stable now :)

PR #676, opened 2020-10-26
Js bundle updates

sbastn:
Apply yarn updates keeping  the `serialize-javascript` library to the correct version.

Closes #672 
Closes #673 
Closes #674


sbastn:
I have turned on grouped dependency updates in Depfu like we have for other repos. I don't think it makes sense to receive individual updates for this repo.

PR #677, opened 2020-10-29
Call out 5-day waiting period to transfer domain

deandre:
I encountered a light "gotcha" situation, in that it may not be
clear to customers that transfers into DNSimple can take up to 5 days.
This change identifies that as a caveat in the docs and links to an old
blog post that gives a bit more context as to why that's the case.

- https://support.dnsimple.com/articles/transferring-domain/
- https://blog.dnsimple.com/2012/01/things-to-know-about-transferring-a-domain/
- https://blog.dnsimple.com/2011/02/preparing-transfer-from-godaddy/

### DIFF
--- a/content/articles/transferring-domain.markdown
+++ b/content/articles/transferring-domain.markdown
@@ -61,7 +61,7 @@ Your domain transfer request will now be issued if there are no missing fields. 
 
 Follow the instructions in the email. We cannot issue the transfer request to the registry without approval of this email.
 
-Once you've authorized the transfer, you may have to **wait up to 5 days your domain transfer to complete.**
+Once you've authorized the transfer, you may have to **wait up to 7 days your domain transfer to complete.**
 
 <info>
 We will only charge your credit card for the transfer once it has completed.
--- a/content/articles/transferring-domain.markdown
+++ b/content/articles/transferring-domain.markdown
@@ -25,7 +25,7 @@ To avoid the possibility of downtime, you should [point the name servers for the
 ## Caveats
 
 - A domain may only be transferred if it was registered more than 60 days ago and has not been transferred within the last 60 days.
-- It can take up to five (5) calendar days for a transfer to complete depending on the domain's Top Level Domain (TLD). To learn more about how to avoid downtime during this period, check out [this article](/articles/before-transferring-domain) for more information.
+- It can take up to seven (7) days for a transfer to complete depending on the domain's Top Level Domain (TLD). To learn more about how to avoid downtime during this period, check out [this article](/articles/before-transferring-domain) for more information.
 - Most registrars will not allow an expired domain to be transferred. However, if the domain expires during the transfer, the registrar may not block the transfer due to the expiration.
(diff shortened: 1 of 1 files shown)


OleMchls:
Hey @deandre,

Thanks so much for jumping right in and providing this PR for us. We've just applied a few tone changes and linked to a more up-to-date article from the support site. We'll have one more 🇺🇸 review and get this shipped right away :shipit:

The transfer procedures for a domain can vary (heavily) on a TLD basis. This is why we need to be more generic in our support articles. For example, TLDs like .co.uk or .be are real-time once all prerequisites are met (auth etc.). Others like all TLDs under ICANN management and .com take up to 5 days.

kayla-broussard (approved):

content/articles/transferring-domain.markdown
```diff
@@ -25,7 +25,7 @@ To avoid the possibility of downtime, you should [point the name servers for the
 ## Caveats
 
 - A domain may only be transferred if it was registered more than 60 days ago and has not been transferred within the last 60 days.
-- It takes up to five (5) days for a transfer to complete and this process generally cannot be expedited, though some registrars may allow explicit  approval. Check out [our post on this topic](https://blog.dnsimple.com/2012/01/things-to-know-about-transferring-a-domain/) for more information.
+- It can take up to five (5) calendar days for a transfer to complete depending on the domains Top Level Domain (TLD). To learn more about how to avoid downtime during this period, check out [this article](/articles/before-transferring-domain) for more information.
```
deandre, line 28:
```suggestion
- It can take up to five (5) calendar days for a transfer to complete depending on the domain's Top Level Domain (TLD). To learn more about how to avoid downtime during this period, check out [this article](/articles/before-transferring-domain) for more information.
```

deandre:
Hey @OleMchls— thanks so much for your review and explanation. I've made a couple more tiny suggestions. :)

content/articles/transferring-domain.markdown
```diff
@@ -60,7 +61,7 @@ Your domain transfer request will now be issued if there are no missing fields.
 
 Follow the instructions in the email. We cannot issue the transfer request to the registry without approval of this email.
 
-Once you've authorized the transfer, you may have to **wait up to 7 days your domain transfer to complete.**
+Once you've authorized the transfer, you may have to **wait up to 5 days your domain transfer to complete.**
```
weppos, line 64: @OleMchls @kayla-broussard FYI there is a reason why we write 7 days. While ICANN says 5 days, systems don't work with second-precision.

Sometimes it's 5 days an 1 hour, sometimes 5 days and many hours (closer to 6). If we factor in time-zones (someone starting at the end of their day, then checking at the beginning of the 5th day), along with delays, the perception is sometimes effectively 6 to 7 days.

We explicitly changed from 5 to 7 in the past due to the number of support requests. We should not revert back.

OleMchls: @weppos I've adjusted this in 31df3de 👍

itsalyse (approved):

OleMchls:
Thanks @deandre for the proposal and @itsalyse for the review :) I'll ship this change to the site right now

PR #690, opened 2020-11-23
Lets encrypt article edits

itsalyse:
Updates Let's Encrypt support article 
- remove mentions of it being "new" (it's 5 years old)
- updates for tone 
- updates for formatting consistency with other support articles


kayla-broussard (approved):
Copy LGTM 👍

PR #696, opened 2020-12-17
Billing information (not email) can be customized on any plan

### LINKED ISSUE #695: Clarify that Professional or higher plan is necessary to customize invoice details
Running into a bit of a back and forth with a customer on a Personal plan who wants to customize the details that are included on their invoices.

I've explained that to do this, it's necessary to customize the billing email first, which is a feature available on higher tiers. Our documention/FAQ is confusing in this respect:

> In the FAQ it says that billing information can be changed by contacting support (https://support.dnsimple.com/articles/billing-settings/#customize-billing-information) - see screenshot. If there is no such option,I find it confusing not knowing that beforehand as I need these informations on the invoice in order to have them registered in accounting.

Raising this issue to make this more apparent in the info tip.

Groove ticket: https://dnsimple.groovehq.com/tickets/103224

deandre:
Edit: following the chat in #695, the ability to change the billing information (company name, address, tax ID) is available for _any_ plan. The billing email can only be changed with Professional and higher, however.

This support article led me to believe that a higher plan is required in order to change such details, but that's not the case. This PR swaps the order of the two sections so as to present a less causal relationship, hopefully clarifying that changing billing details is a go across all tiers.

~`@jacegu` Sarah suggested I ping you to chat about this, hope you don't mind if we take this async. More context is in issue #695, but our basic question is whether it's even possible to add customized billing information to invoices where an account is on a Personal plan.~

~If it's not at all possible, we should definitely clarify such in the documentation, which is what this PR does.~

~If it is possible, is it something we'd want to do (maybe on a case by case basis) or should we encourage users to upgrade to an appropriate plan?~

Closes #695.

### DIFF
--- a/content/articles/billing-settings.markdown
+++ b/content/articles/billing-settings.markdown
@@ -19,7 +19,7 @@ Until you provide us with further information, your invoices will reflect the em
 ![Example Invoice Without Billing Information](/files/account-billing-settings-invoice-1.png)
 
 
-## Customize billing information
+## Customizing billing information
 
 You can replace the billing details with custom information on all your invoices:
 
@@ -50,7 +50,7 @@ You can modify this information any time, as many times as you need to. The new 
 
 All future invoices will display the information you provided.
 
-## Change the billing email
+## Changing the billing email
 
 We can deliver your invoices to a different email address from the one you use to manage your domains.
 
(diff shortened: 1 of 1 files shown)


itsalyse (approved):

SarahCaminiti1 (approved):
LGTM!

PR #702, opened 2020-12-28
Update docs to make ICANN-lock for transfers between accounts more apparent

deandre:
We've recently found that the 60-day lock also applies to domains transferred between DNSimple accounts. This came about because one reseller was waiving the lock while the other wasn't.

This change makes the lock a bit more explicit in a couple of relevant articles.

### References

- https://dnsimple.groovehq.com/tickets/103787

### DIFF
--- a/content/articles/transferring-domain-away.markdown
+++ b/content/articles/transferring-domain-away.markdown
@@ -21,6 +21,9 @@ This article describes the procedure to transfer a domain registered with DNSimp
 
 <warning>
 Once a domain transfer is started, you won't be able to change the name servers on the domain. If you want to change your name servers, you can do so after requesting the code. Once you provide the transfer code to the gaining registrar, you cannot make adjustments.
+<br /><br />
+    
+If you've recently had a domain transferred to your DNSimple account from another, you won't be able to transfer the domain away for 60 days due to an [ICAAN-imposed lock after change of registrant](/articles/icann-60-day-lock-registrant-change/).
 </warning>
 
 To transfer a domain to another registrar, the standard procedure requires you to:
--- a/content/articles/transferring-domain-away.markdown
+++ b/content/articles/transferring-domain-away.markdown
@@ -21,9 +21,6 @@ This article describes the procedure to transfer a domain registered with DNSimp
 
 <warning>
 Once a domain transfer is started, you won't be able to change the name servers on the domain. If you want to change your name servers, you can do so after requesting the code. Once you provide the transfer code to the gaining registrar, you cannot make adjustments.
-<br /><br />
-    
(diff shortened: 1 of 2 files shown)


weppos:
@deandre the lock is applied ONLY if the transfer result in a registrant change. Regardless whether it's intra-accounts or not. The lock also applies if you just change contact even without transferring to any other account.

I think the changes in this PR would make things a bit more confusing as it seems to imply special rules between account transfers.

If we want to clarify anything, I'd rather prefer we state exactly what ICANN says. Which is any change of contact would result into 60-day transfer lock. Therefore, in our case, if the customer is receiving a domain from another account and a new contact is applied, the lock will become effective. Likewise, any explicit change to the contact resulting in a registrar change will activate the lock.

As it is phrased right now, it is misleading.

content/articles/transferring-domain-away.markdown
```diff
@@ -21,6 +21,9 @@ This article describes the procedure to transfer a domain registered with DNSimp
 
 <warning>
 Once a domain transfer is started, you won't be able to change the name servers on the domain. If you want to change your name servers, you can do so after requesting the code. Once you provide the transfer code to the gaining registrar, you cannot make adjustments.
+<br /><br />
+    
+If you've recently had a domain transferred to your DNSimple account from another, you won't be able to transfer the domain away for 60 days due to an [ICAAN-imposed lock after change of registrant](/articles/icann-60-day-lock-registrant-change/).
```
weppos, line 26: FYI this is misleading. See my comment. It's not true that if you received a domain from another account it is automatically locked. It depends on the TLD (whether it's an ICANN-regulated TLD or not), and more importantl whether a registrant change (change of contact) was applied.

content/articles/transferring-domain-between-accounts.markdown
```diff
@@ -23,6 +23,8 @@ Working with a Reseller? Follow the steps outlined [here](#accepting-a-transfer)
 
 <warning>
 Once the transfer is accepted, you'll no longer be able to manage the domain.
+<br /><br />
+When transferring a domain to another DNSimple account, the domain's registrant information will be updated to reflect the new account. This may result in the domain being [locked from further transfers for 60 days](/articles/icann-60-day-lock-registrant-change/).
```
deandre, line 27: @weppos Thanks for your comments—I understand better now. Does this change make more sense?

SarahCaminiti1: LGTM

SarahCaminiti1 (approved):

itsalyse (approved):
copy looks good to me, pending @weppos's approval 👍

weppos (approved):
Much better! Thanks 👏

PR #711, opened 2021-01-04
Allow margin for p in markers

weppos:
Otherwise we are forced to use HTML as in https://github.com/dnsimple/dnsimple-support/pull/702

This rule was added in https://github.com/dnsimple/dnsimple-support/pull/451 but I cannot find exactly why we needed it (it is incorporated in a bunch of other changes).

Also this rule doesn't exist on the developer site, and I'm confident the style for the markers could/should be the same
https://github.com/dnsimple/dnsimple-developer/blob/master/content/assets/css/_application.scss
In fact, should we sync the marker definition between these two projects?

Screenshot 2021-01-04 at 15 22 44

### DIFF
--- a/content/assets/css/_application.scss
+++ b/content/assets/css/_application.scss
@@ -24,9 +24,6 @@
   h4 {
     margin-top: 0px;
   }
-  p {
-    margin: 0;
-  }
 
   &:before {
     display: block;
--- a/content/assets/css/_application.scss
+++ b/content/assets/css/_application.scss
@@ -25,6 +25,12 @@
     margin-top: 0px;
   }
 
+  p {
+    &:last-child {
+      margin-bottom: 0;
+    }
+  }
+
   &:before {
     display: block;
     float: left;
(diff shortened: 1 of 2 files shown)


content/assets/css/_application.scss
```diff
@@ -24,9 +24,6 @@
   h4 {
     margin-top: 0px;
   }
```
dallasread, line 26:
```suggestion
  }
  
  :last-child {
    margin: 0;
  }
```

dallasread, line 26:
```suggestion
  }
  
  :last-child {
    margin-bottom: 0;
  }
```

dallasread (approved):
👍 LGTM. I left a suggestion that ¿_may_? align more with the intention of the existing code.

It had not occured to me that we were trying to keep the repos somewhat in sync. 😬

content/assets/css/_application.scss
sbastn: ✅ applied suggestion

sbastn: Correct, applied suggestion

sbastn (approved):

weppos:
Thanks.

PR #713, opened 2021-01-05
Updates bundler version in lock file

DXTimer:
Locks the bundler version to latest. It also updates the previously revert bundler API change as the old method is deprecated.


sbastn (approved):
Works as expected! This guide was useful for me to upgrade Bundler: https://bundler.io/guides/bundler_2_upgrade.html Note that I had to do it on another folder that is not this repo.

PR #715, opened 2021-01-06
Update CAA instructions for Sectigo and wildcard

weppos:
Comodo is now Sectigo. I also moved the warning block about CAA closer to where it can be searched (in the troubleshooting).

We are probably missing some information in the instructions to purchase a new certificate. But for now, I think this is a better fit than having Sectigo-specific instructions buried into a generic CAA article.


SarahCaminiti1 (approved):

deandre (approved):
LGTM!

PR #721, opened 2021-01-14
Add invoice explanation

deandre:
As reported in "The Breakdown" (Sarah's newsletter), 38 out of 237 tickets for the sampled time period were from customers who had confusion after looking at their invoice.

This PR introduces a new article that explains DNSimple's invoices. It also updates existing articles with information about what extra domain zones are and which billing fields can be customized on an invoice.

@jacegu would you mind doing a first round review? I'd like to lean on your billing expertise. I'll then pull in Sarah and Alyse.


content/articles/understanding-invoice.markdown
```diff
@@ -0,0 +1,73 @@
+---
+title: Understanding Your DNSimple Invoice
+excerpt: A detailed explanation of DNsimple invoices.
```
jacegu, line 3:
```suggestion
excerpt: A detailed explanation of DNSimple invoices.
```

content/articles/understanding-invoice.markdown
```diff
@@ -0,0 +1,73 @@
...
+
+### Recurring charges
+
+- DNSimple subscription
+- [Email forwards](/articles/email-forwarding/)
+- [Extra domain zones](/articles/dns-hosting/#extra-zones)
+
+![Recurring charges](/files/recurring-charges.png)
+
+### One-time charges
```
jacegu, line 51: We frequently refer to these 1 time charges as "purchases". Just bringing this up in case we want to also have this term in this article.

jacegu (approved):
Changes and new article LGTM. Thanks for putting this together @deandre

itsalyse (approved):

Issue #725, opened 2021-01-20
Document the need to delete DS records when switching to DNSimple upon transfer


weppos:
When you transfer a domain to DNSimple, we sync the DS records and we do not alter the delegation.

However, if the user changes delegation to DNSimple and the zone is not signed, the DS record will conflict and cause the domain to stop resolving.

1. https://support.dnsimple.com/articles/transferring-domain/#after-the-transfer needs to be expanded. Where we instruct the customer to change the delegation, we must first add a point to review the DS records and remove them if the zone is not signed at DNSimple
2. https://support.dnsimple.com/articles/before-transferring-domain/ has a mention about ensuring there are no DS records, but it may use some additional warning formatting that explains existing DS record may cause the domain to stop resolving if you later on change delegation

ggalmazor:
Adding a link to an important insight we need to somehow explain too: https://github.com/dnsimple/dnsimple-support/issues/657#issuecomment-763807209

ggalmazor:
Maybe update this one too? https://dnsimple.com/how-to-transfer-without-downtime

aeden:
> Maybe update this one too? https://dnsimple.com/how-to-transfer-without-downtime

Yes.

deandre:
@aeden both PRs have been merged and this is now complete.

PR #731, opened 2021-02-01
Add more info about DS records when changing delegation

deandre:
Related to https://github.com/dnsimple/dnsimple-support/issues/725

### DIFF
--- a/content/articles/before-transferring-domain.markdown
+++ b/content/articles/before-transferring-domain.markdown
@@ -18,7 +18,7 @@ Transferring domain names from one registrar to another is often something that 
 
 ## Ensure DNSSEC is disabled
 
-If you are currently using DNSSEC, then you must disable it before changing the name servers.
+If you are currently using DNSSEC, make sure to disable it at your registrar before changing the name servers.
 
 Then you must [remove the current DS record](/articles/ds-records-changing-dns) before transferring your domain away from your current provider.
 
--- a/content/articles/before-transferring-domain.markdown
+++ b/content/articles/before-transferring-domain.markdown
@@ -18,7 +18,9 @@ Transferring domain names from one registrar to another is often something that 
 
 ## Ensure DS records are removed
 
-If you are currently using DNSSEC, then you must [enable it for the domain](/articles/dnssec/#enabling-dnssec) in DNSimple so that we sign the zone. Doing so will allow you to safely remove DS records from your current provider.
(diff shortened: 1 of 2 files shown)


content/articles/before-transferring-domain.markdown
```diff
@@ -18,7 +18,7 @@ Transferring domain names from one registrar to another is often something that
 
 ## Ensure the DS record is removed
```
ggalmazor, line 19:
```suggestion
## Ensure DS records are removed
```
(we can't assume there will only be one of them)

content/articles/before-transferring-domain.markdown
```diff
@@ -18,7 +18,7 @@ Transferring domain names from one registrar to another is often something that
...
 
-If you are currently using DNSSEC, then you must [remove the current DS record](/articles/ds-records-changing-dns) before transferring to our name servers.
+If you are currently using DNSSEC, then you must [enable it for the domain](/articles/dnssec/#enabling-dnssec) in DNSimple so that we sign the zone. Doing so will allow you to safely remove DS records from your current provider.
```
ggalmazor, line 21:
```suggestion
If you are currently using DNSSEC, then you must [enable it for the domain](/articles/dnssec/#enabling-dnssec) in DNSimple so that we sign the zone too. Doing so will allow you to safely remove DS records from your current provider.
```

content/articles/transferring-domain.markdown
```diff
@@ -96,3 +96,7 @@ We suggest pointing [your domain to our name servers before the transfer](/artic
 Most transferred domains will be extended the minimum required extension period. For example, `.com` domains will always be extended one year when they are transferred.
 
 The price of this extension is included in the transfer fee.
+
+### DS records
+
+Before delegating your domain to DNSimple make sure that you first [enable DNSSEC for the domain](/articles/dnssec/#enabling-dnssec) in order to safely remove the DS records from your current provider.
```
ggalmazor, line 102: I think we also need to make this suggestion optional only for those customers using DNSSEC in their old providers.

ggalmazor (commented):
👍 

The bottom line change about enabling DNSSEC at DNSimple before starting the transfer/delegation change feels like the best way to make it safe for our customers to remove DS records provisioned by their other DNS providers.

deandre:
Thanks for the review, @ggalmazor! Going to ask @aeden to take a look now.

ggalmazor (approved):
LGTM!

itsalyse (approved):

PR #733, opened 2021-02-09
Change: update screenshot

ggalmazor:
This PR updates the screenshot at https://support.dnsimple.com/articles/srv-record/#add-an-srv-record to show the current UI.

I got complaints from a customer that wasn't able to follow the article due to the difference between the screenshot and the form we're currently showing users in the app.

I decided to leave out the region selector and the preview (which is not working properly at the moment) because it feels like it would deserve to expand the article.

This would be the result:

image


sbastn (approved):

dallasread (approved):
👍  ⚡  🥇

PR #735, opened 2021-02-10
Clarify that push wants Account emails

deandre:
As disussed in https://github.com/dnsimple/dnsimple-app/pull/15603, it could be clearer that push wants Account emails. This PR adds a bit more detail about that.


deandre:
Once the changes in the referenced PR are merged we'll also need to update the screenshot in this article.

itsalyse (approved):

SarahCaminiti1 (approved):
This is perfect

PR #740, opened 2021-02-24
Add 'What is a name server?' article

deandre:
This PR adds an article that explains what name servers are in simple terms. In the interest of simplicity, I use the script from the video explanation, though this can be expanded in future iterations.

Closes https://github.com/dnsimple/dnsimple-business/issues/1234.


deandre:
@laetitiaeden would you mind giving this a review? Is this aligned with what you have in mind?

I'm a bit concerned about duplicating the step-by-step instructions in this article as separate articles address those.

Here's the preview: https://deploy-preview-740--dnsimple-support.netlify.app/articles/what-is-a-nameserver/

laetitiaeden:
I think you should only put the link not the full article specially if you have other interesting items in the article like the FAQ and the glossary

itsalyse (approved):

itsalyse (approved):

laetitiaeden:
Ok I apologized but after looking at the instructions, I am changing my mind.  I think we should keep them they are not duplicate, they are actually specific to this task of changing your name server to DNSimple and they are more details than the existing one.

deandre:
> Ok I apologized but after looking at the instructions, I am changing my mind. I think we should keep them they are not duplicate, they are actually specific to this task of changing your name server to DNSimple @SarahCaminiti1 created them specifically they are more detailed.

@laetitiaeden I'm partial to updating the existing article with these more specific instructions and linking to that article from here, instead. What are your thoughts? /cc @SarahCaminiti1

laetitiaeden:
@deandre you choose what you think is the best approach.  Link to the article if you are more comfortable with it.

laetitiaeden (approved):

aeden:
> > Ok I apologized but after looking at the instructions, I am changing my mind. I think we should keep them they are not duplicate, they are actually specific to this task of changing your name server to DNSimple @SarahCaminiti1 created them specifically they are more detailed.
> 
> @laetitiaeden I'm partial to updating the existing article with these more specific instructions and linking to that article from here, instead. What are your thoughts? /cc @SarahCaminiti1

The only thing I ask is that you link now, publish this, and then make a separate PR to make adjustments to other articles in the future.

PR #742, opened 2021-02-24
Layout changes on index page

sbastn:
While looking at our support site, I found a few layout problems that I want to address with this PR.

- The list of items below the "Looking for something else?" section is growing and becoming harder to read.
- The important links to the starting guide or personal plan guide are hidden in the middle of the page. 
- Video at the top of the page is not sized properly
- Image for "need help" is not sized properly

## Before

Screenshot_2021-02-24 DNSimple Help - DNSimple Help

## After
Screenshot_2021-02-24 DNSimple Help - DNSimple Help(3)

### DIFF
--- a/content/index.html
+++ b/content/index.html
@@ -4,7 +4,12 @@ excerpt: Home of the DNSimple domain management help documentation and tutorials
 ---
 
 <h1>We're here to help</h1>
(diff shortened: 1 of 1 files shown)


dallasread (approved):
👍  This a great update. Thanks!

laetitiaeden (approved):
@sbastn we should put on the calendar to look at the page again in one month. Let's take one minute to check the mobile view to make sure it all looks good.

PR #768, opened 2021-05-12
Update the instructions that the user needs to follow after setting up secondary dns as secondary

ggalmazor:
This PR introduces new steps to complete the Secondary DNS as Secondary setup while we work on a more streamlined process for our users.

- [ ] POST: update links to the changed pages at the DNSimple app


deandre (approved):
LGTM!

PR #784, opened 2021-06-11
DNSimple as secondary DNS video walk-through

sbastn:
Adds the video walk-through to the top of the support article for DNSimple as Secondary DNS provider.

Screen Shot 2021-06-11 at 12 00 31 PM

### DIFF
--- a/content/articles/secondary-dns-dnsimple-as-secondary.markdown
+++ b/content/articles/secondary-dns-dnsimple-as-secondary.markdown
@@ -14,15 +14,7 @@ For an overview of Secondary DNS, have a look at [our introduction article](/art
 * TOC
 {:toc}
 
----
-
-<warning>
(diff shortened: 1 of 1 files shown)


laetitiaeden:
@sbastn should we mention that the video is with DNS made easy or it doesn't matter?

dallasread (approved):
Thanks for doing this – it looks great. 

Because the video and article are geared to be a generic overview, I don't think there is a need to mention that it is DNS Made Easy. If we do mention it, we also have to clarify that it can be done with any other secondary-supporting provider.

laetitiaeden (approved):

sbastn:
> @sbastn should we mention that the video is with DNS made easy or it doesn't matter?

I personally think that it doesn't matter. You'll still need to perform the same steps regardless of the provider that you use.

PR #796, opened 2021-07-14
Support Widget: Search by URL

dallasread:
The support widget's default state is showing no results in some cases. This is because `dnsimple-app` is expecting to be able to supply a URL to search by. This PR will unlock the ability for `dnsimple-app` to specify a URL on a per-page basis. I took the opportunity to extract a couple methods for clarity.

To review: 
- Search for `/articles/whois-privacy/` – a _single_ item should appear.


sbastn (approved):
👏

PR #797, opened 2021-07-15
Add AU article

deandre:
AU is proving one of those complicated TLDs. We should document some of the requirements and processes in a public-facing article.

### DIFF
--- a/content/articles/domains-au.markdown
+++ b/content/articles/domains-au.markdown
@@ -61,7 +61,7 @@ A valid Australian Business Number (ABN) is required to register a `.ORG.AU` dom
 
 ## Renewals
 
-`.AU` domains must be renewed no later than 10 days before the expiration date. Automatic and manual renewals are permitted as long as the domain is less than 90 days away from expiry. If the domain name is not renewed by the deadline, it will enter redemption. The domain name is removed from the DNS and cannot be updated. This period lasts for 31 days; during this period, the domain name can be recovered for an additional fee.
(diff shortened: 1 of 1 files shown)


content/articles/domains-au.markdown
```diff
@@ -0,0 +1,74 @@
...
+- A charity operating in Australia, as defined in the Registrant's constitution or other documents of incorporation.
+- A non-profit organization operating in Australia, as defined in the Registrant's constitution or other documents of incorporation.
+
+<warning>
+A valid Australian Business Number (ABN) is required to register a `.ORG.AU` domain name. Australian Company Numbers and Trademark Numbers are not acceptable.
+</warning>
+
+## Renewals
+
+`.AU` domains must be renewed no later than 10 days before the expiration date. Automatic and manual renewals are permitted as long as the domain is less than 90 days away from expiry. If the domain name is not renewed by the deadline, it will enter [redemption](/articles/what-happens-when-domain-expires/#redemption-period). The domain name is removed from the DNS and cannot be updated. This period lasts for 31 days; during this period, the domain name can be recovered for an additional fee.
```
weppos, line 64: > Automatic and manual renewals are permitted as long as the domain is less than 90 days away from expiry.

Unfortunately, we are limited by our current partner and we can't support manual renewal.

deandre: @weppos wait, these tlds are auto renew only?

weppos: At the time being yes, .AU is auto-renewal only. You cannot manually renew it.

deandre: Ahh, I see. Should the suffixes be updated? The name tool indicates otherwise.

weppos: After review, my comment is no longer valid.

itsalyse (approved):

SarahCaminiti1 (approved):

PR #821, opened 2021-11-17
Adds ICANN domain expiration email article

deandre:


SarahCaminiti1:
The emails aren't coming from ICANN, they're coming from Enom or Hexonet. I would also specify where you say that they could get it up to 3 times that this is for every domain. Please also include the Hexonet email address

deandre:
> The emails aren't coming from ICANN, they're coming from Enom or Hexonet. I would also specify where you say that they could get it up to 3 times that this is for every domain. Please also include the Hexonet email address

What's the email address for Hexonet? All the examples folks have forwarded are only from name-services.

deandre:
> > The emails aren't coming from ICANN, they're coming from Enom or Hexonet. I would also specify where you say that they could get it up to 3 times that this is for every domain. Please also include the Hexonet email address
> 
> What's the email address for Hexonet? All the examples folks have forwarded are only from name-services.

@weppos Do you happen to know Hexonet's email address for ERRP expiration notices — whether it differs from `donotreply@name-services.com`?

weppos:
> @weppos Do you happen to know Hexonet's email address for ERRP expiration notices — whether it differs from `donotreply@name-services.com`?

I think it is `no-reply@domainrenewals.ispapi.net`

deandre:
> > @weppos Do you happen to know Hexonet's email address for ERRP expiration notices — whether it differs from `donotreply@name-services.com`?
> 
> I think it is `no-reply@domainrenewals.ispapi.net`

That's it, thank you! Was also able to find examples to screenshot in Groove. :)

itsalyse (approved):

PR #825, opened 2021-12-07
Adds some text to the Managing DS records section

ecomba:
There is a lack of symetry in the DNSSEC page. To rememdy this I've added some text to the manage ds records section.

At some point we could have a better look at the content in our support page (from a high level view).


itsalyse (approved):

dallasread (approved):
LGTM! 🥇

PR #838, opened 2022-01-13
Adds contribution guidelines

deandre:
Since this repo is open source and the team is growing, there should be contribution guidelines. This PR adds a first round — we should refine these guidelines in the future and probably place them in their own file.


weppos:
I suggest to reference https://support.dnsimple.com/articles/writing-guide/, which is the style guideline to follow when writing/editing articles.

SarahCaminiti1 (approved):

PR #876, opened 2022-03-02
Transfer away expired domain

deandre:
Closes https://github.com/dnsimple/dnsimple-cs/issues/11.

### DIFF
--- a/content/articles/what-happens-when-domain-expires.markdown
+++ b/content/articles/what-happens-when-domain-expires.markdown
@@ -21,13 +21,13 @@ Certain domain names are [auto-renew only](/articles/domain-auto-renewal) and ca
 
 After a domain expires, it enters a grace period. In this period, the domain is disabled but it's still possible to renew it from the DNSimple account at no extra cost.
 
-Our registrar partners may offer a non-guaranteed 29-day grace period for renewal for most generic TLDs (gTLDs). There is no guaranteed grace period. Once a domain expires, our registrar partners may dispose of it at any time according to the registration agreement. After the grace period elapses, the domain may go into a redemption grace period.
+Our registrar partners may offer a non-guaranteed 29-day grace period for renewal for most generic top-level domains (gTLDs). There is no guaranteed grace period. Once a domain expires, our registrar partners may dispose of it at any time according to the registration agreement. After the grace period elapses, the domain may go into a redemption grace period.
 
 Renewal years are always added to the end of the current registration period (i.e., the new expiration date is the same whether you renew two months early or the day before expiration), so you will never lose time on your domain by renewing early.
 
 #### Transferring away expired domain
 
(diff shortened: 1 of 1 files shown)


itsalyse (approved):

weppos:
@deandre do you think it is worth to make further distinction and/or clarification between ccTDLs and gTLDs? While this may be too much to interpret, I also cannot unsee the fact we're sometimes mixing up topics in a way that could be misleading.

Allow me to explain.

> Our registrar partners may offer a non-guaranteed 29-day grace period for renewal for most TLDs. There is no guaranteed grace period. Once a domain expires, our registar partners may dispose of it at any time according to the registration agreement. After the grace period elapses, the domain may go into a redemption grace period.

I think this is good. However, the grace period only applies consistently for gTLDs as this is an ICANN requirement. ccTLDs are not bounded to this requirement, therefore the outcome may varies significantly. Would is be worth making a distinction here between the TLD type? Not critical, but perhaps helpful.

> #### Transferring away expired domain
> 
> ICANN's rules and regulations permit expired domains to be transferred to another registrar, provided the domain is still in its [grace period](#grace-period). Transfers must complete before the domain falls into its [redemption period](#redemption-period).

This is actually where I think it's important to make a small distinction, We're talking about a generic "domains", then we cite ICANN. It may not be obvious to most readers that ICANN rules only apply to gTLD, therefore this paragraph only applies to gTLDs.

I think if we want to add this paragraph, we must clarify this is valid ONLY for gTLDs, or we may have customers reaching out expecting this rule would apply e.g. to their .IT or .CA domain.

content/articles/what-happens-when-domain-expires.markdown
```diff
@@ -21,10 +21,16 @@ Certain domain names are [auto-renew only](/articles/domain-auto-renewal) and ca
 
 After a domain expires, it enters a grace period. In this period, the domain is disabled but it's still possible to renew it from the DNSimple account at no extra cost.
 
-Currently, eNom – our registrar - offers a non-guaranteed 29-day grace period for renewal for most TLDs. There is no guaranteed grace period. Once a domain expires, eNom may dispose of it at any time according to the registration agreement. After the grace period elapses, the domain may go into a redemption grace period.
+Our registrar partners may offer a non-guaranteed 29-day grace period for renewal for most TLDs. There is no guaranteed grace period. Once a domain expires, our registar partners may dispose of it at any time according to the registration agreement. After the grace period elapses, the domain may go into a redemption grace period.
```
weppos, line 24:
```suggestion
Our registrar partners may offer a non-guaranteed 29-day grace period for renewal for most TLDs. There is no guaranteed grace period. Once a domain expires, our registrar partners may dispose of it at any time according to the registration agreement. After the grace period elapses, the domain may go into a redemption grace period.
```

weppos (approved):

weppos:
> @weppos really good feedback, thank you. i've made some edits for clarity — would you take a look? the document itself is already structured to differentiate between icann/non-icann TLDs, but no harm in making things painfully clear.

Much better.

SarahCaminiti1 (approved):

PR #879, opened 2022-03-07
Netlify connector

sbastn:
Adds new article for the Netlify connector.

### DIFF
--- a/content/articles/netlify-connector.markdown
+++ b/content/articles/netlify-connector.markdown
@@ -50,5 +50,5 @@ In this tab, you can view the metadata of the connection, or disconnect the Netl
 
 ## Connection monitoring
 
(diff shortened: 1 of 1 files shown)


aeden (commented):
I've added @itsalyse to perform copy editing, and will review the text in full when she completes her first pass.

I also think it'd be good if you could update the Netlify Service article to indicate the service is deprecated and that the connector should be used instead.

dallasread (approved):
Nice work! 

Can we remove the video for this PR and it when it is ready. WDYT? 😄

itsalyse (approved):

aeden:
I've added a section about connection monitoring. @itsalyse and @dallasread could you please re-review with that section.

itsalyse (approved):

PR #907, opened 2022-06-01
Remove mention contacts or having to choose a contact from certificate related processes

ggalmazor:
Belongs to https://github.com/dnsimple/dnsimple-business/issues/1500

Our system no longer requires our customers to provide a contact to purchase certificates. We now use a stub contact instead.

This PR updates articles about requesting certificates to remove mentions about having to choose a contact or how contacts are used during the process.


content/articles/ordering-lets-encrypt-certificate.markdown
```diff
@@ -49,26 +49,10 @@ The order is the first step into getting an SSL certificate. It will create an S
...
-
-Other plans provide the ability to configure only one name on your Let's Encrypt certificate:
-
-![Let's Encrypt with single name configuration](/files/lets-encrypt-new-names.png)
-
-On some plans we do not allow the configuration of a name but instead configure the certificate to work only on the apex and www-subdomain of the zone:
-
-![Let's Encrypt with fixed names](/files/lets-encrypt-new-fixed.png)
-
-
```
jacegu, line 71: I'm doubtful about this removal... I think it is useful so customers can visually tell the difference when setting up LE certs with different name configurations.

content/articles/what-is-csr.markdown
```diff
@@ -12,7 +12,7 @@ The **Certificate Signing Request** (also **CSR** or **certification request**)
 The CSR contains information that will be included in your certificate, such as the [common name](/articles/what-is-common-name) and the company/owner details. It also contains the public key that will be embedded in the certificate.
 
 <info>
-We automatically generate the CSR for you using the information from the contact associated with the certificate along with an unique public/private encryption key pair. This works for the majority of cases, including hosting platforms such as Heroku or Amazon.
+We automatically generate the CSR for you using an unique public/private encryption key pair. This works for the majority of cases, including hosting platforms such as Heroku or Amazon.
```
jacegu, line 15:
```suggestion
We automatically generate the CSR for you using a unique public/private encryption key pair. This works for the majority of cases, including hosting platforms such as Heroku or Amazon.
```

For some reason this sounded wrong to me (even as a 🇪🇸). Funny enough it was highlighted as a spelling error. [I also looked it up online](https://livexp.com/blog/a-unique-or-an-unique-which-article-is-correct-in-this-case/)

jacegu (commented):
Left a couple comments that I think are worth addressing before merging this.

content/articles/what-is-csr.markdown
ggalmazor: Thanks! I agree :)

content/articles/ordering-lets-encrypt-certificate.markdown
ggalmazor: I'm really skeptical about this... the purpose of the removed section is not clear to me.

If it's about explaining what's the difference between plans regarding ordering LE certs, then I'd argue that the screenshots don't provide any valuable information. A simple bullet-point list would be much more efficient for that.

However, the pricing page does a better job comparing plans... Maybe we could replace the whole section with a message like "Check out our [pricing page](https://dnsimple.com/pricing) to understand limitations on your plan regarding LetsEncrypt certificates"

If we maintain the current section, then we're duplicating information in a way that is hard for us to maintain and (IMO) doesn't provide much value to our customers.

jacegu: 🤷‍♂️ I have no strong opinion on this. So if you think it's worth removing, I'm ok with it.

jacegu (approved):

PR #933, opened 2022-08-11
Adjust the name of the commercial SSL certificate products we sell

ggalmazor:
Belongs to https://github.com/dnsimple/dnsimple-business/issues/1539

The product we sell now is called `PositiveSSL`

This PR **MUST NOT** be released before we're ready to enable globally the Sectigo integration


nestorsalceda (approved):

ggalmazor:
@jacegu @nestorsalceda I think we should probably roll this one out. WDYT?

jacegu:
> @jacegu @nestorsalceda I think we should probably roll this one out. WDYT?

Yes. Doing that now.

PR #941, opened 2022-09-20
Remove section about existing name with DNSimple

OleMchls:
We don't support sponsoring namers anymore so we can remove that section already.


OleMchls:
Verified using the deploy preview, moving forward to unblock upcoming changes.

PR #942, opened 2022-09-20
Remove AUda direct info

OleMchls:
With the stop of .AU direct applications on Sept 20th on 23:59 UTC we can remove that info from our support page as well.


juankuquintana (approved):
👍

PR #943, opened 2022-09-26
Improve layout

sbastn:
Before I start working on https://github.com/dnsimple/dnsimple-marketing/issues/544, I want to propose an improvement to the layout for this site. If approved, I'll be happy to deploy it also on the developer site. This is a spike and should be treated as such.

# Current layout issues

## A narrow container

The width of the main container is 1008px wide, which is narrow for today's standards. Line length on articles is good, but this width makes it harder to add new elements like I'm trying to do on https://github.com/dnsimple/dnsimple-marketing/issues/544
Screenshot 2022-09-26 at 16 48 52

## Not really responsive

Navigation is poor. The categories shift from the left of the page to the bottom on smaller screen which is not a standard experience on mobile.

Screenshot 2022-09-26 at 16 53 36

## Confusing navigation

Articles that belong to different categories are good! But this feels like a bug and doesn't help the user.

Screenshot 2022-09-26 at 15 34 55

# Proposal

- A responsive sidebar. It appears as fixed on mobile and tucked away in mobile.
- Do not highlight the different categories that an article belong to on the sidebar.

These are small changes, but I think they improve the experience with the support site. You can browse the full site on Netlify's preview link.

## QA

- Open https://deploy-preview-943--dnsimple-support.netlify.app/
- Find an article
- Use the sidebar
- Resize your viewport
- Toggle the sidebar
- Click away from the sidebar

### DIFF
--- a/content/assets/css/_application.scss
+++ b/content/assets/css/_application.scss
@@ -202,18 +202,3 @@
   color: $accent-color;
   border-left: 5px solid $accent-color;
 }
-
-.articles-list {
-  list-style-type: none;
-  padding: 0;
-}
-
-.articles-list-item {
-  padding: 5px 0;
-
-  &:before {
-    margin-right: 10px;
-    color: #dddddd;
-    @include font-awesome("\f054");
-  }
-}
--- a/content/assets/css/_template.scss
+++ b/content/assets/css/_template.scss
@@ -28,11 +28,6 @@ body {
   background-size: 900px 10px;
 }
 
-.sticky {
-  position: sticky;
-  top: 0;
-}
-
 .sidebar {
   display: none;
   z-index: 100;
--- a/content/assets/css/_template.scss
+++ b/content/assets/css/_template.scss
@@ -27,3 +27,27 @@ body {
   background-repeat: repeat-x;
   background-size: 900px 10px;
 }
+
+.sticky {
+  position: sticky;
+  top: 0;
+}
+
+.sidebar {
+  display: none;
+  z-index: 100;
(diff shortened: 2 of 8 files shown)


dallasread:
I'm generally on board with this. A few things I noticed:

- Can we decrease the page padding when viewing on a mobile?
- Can we have a little space to the right of the mobile hamburger?
- Can we have some visual marker for the items in the sidebar (eg. like a `•` or `-`)
- Can we have the sidebar overflow for when window that aren't very tall?
- Just underneath the post title, what do you think about linking to it's categories? Or even at the bottom of the post?

SarahCaminiti1 (commented):
Love that this is happening!!!!

Looking at the preview I have a few things

- Love the side bar, but the gray color seems too light. Is there another direction we could take the color? If we want to make it more modern and inviting, could we add more color to it?
- Not centering the home page looks like a mistake. Is there a reason everything is to the left?
- More of an organizing point, but when I go to a category like Domains, the list is looonnnngggg and it's overwhelming. Could we have subcategories?
- We have the great characters from the comics, could we make the support page more inviting by incorporating the color and playfulness here? Most folks that are using the support page are overwhelmed, having it be a fun and less daunting space could make it more accessible? 

If you want to hop on a call and talk about it more let me know

sbastn:
Closing spike. Will come back to this later.

dallasread:
I came across this similar design. I love seeing the headings TOC on the right.

https://docs.airplane.dev/self-hosting/aws

PR #944, opened 2022-09-28
Sub categories for categories pages

sbastn:
This PR adds sub-categories for category pages.

## Before
Screenshot 2022-09-28 at 15 08 07

## After
Screenshot 2022-09-28 at 15 08 21

## How it works

1. Each category can optionally have a YAML file in the `categories` folder. If there is no YAML file, the category page still displays all the articles, but all they appear all under the "Other" sub-category.

3. The YAML name should match the category name. For example, if the category is "Secondary DNS", the YAML file will be "secondary_dns.yaml"

4. The YAML file doesn't need to list all the articles in the category. If an article is not listed, it will be added to the "Other" subcategory. The "Other" subcategory is built on the fly. There is no need to define it.

## QA

1. Open the preview netlify site: https://deploy-preview-944--dnsimple-support.netlify.app/
2. Navigate to the category pages
3. You should see sub-categories.

I sub-categorized to the best of my knowledge. In doing this, I noticed that some articles were categorized incorrectly, or are in multiple categories when they shouldn't. I didn't update any article's categories, but it should probably be done next. Those articles that weren't a good fit were left in the "Other" section that gets automatically created if the article is not in the YAML for the category page.

I think all of the "XXXX and DNSimple" should be rolled into one category page now. Maybe "Integrations" is the correct place for it. I'd like to reach an agreement on this.

Screenshot 2022-09-28 at 15 27 18

## Next steps

1. Category pages contain the correct articles
2. Articles with wrong categories are updated
3. Agree on moving or leaving as-is the "XXXX and DNSimple"


aeden:
It's really nice, @sbastn . I agree that the last 3 menu items could be moved into Integrations, but it's not strictly necessary if you'd rather wait.

aeden (approved):

SarahCaminiti1 (approved):
I AM OBSESSED

PR #948, opened 2022-10-04
Update support article on our level of support of ECC signed SSL certificates

DXTimer:
Update support article on our level of support of ECC signed SSL certificates:

* We now support ECC certs via custom CSR
* List supported curve functions


ggalmazor (approved):

nestorsalceda (approved):
Thank you!

PR #949, opened 2022-10-04
Add instructions for .DAY

OleMchls:
We need to provide some additional info when providing .DAY domains.

### DIFF
--- a/content/articles/domains-day.markdown
+++ b/content/articles/domains-day.markdown
@@ -19,6 +19,6 @@ This article explains the requirements and special procedures for .DAY domain na
 
 .DAY is a secured namespace, meaning that websites using a .DAY domain require an SSL certificate to work; .DAY websites can only be accessed by web browsers using HTTPS through an encrypted and secured connection.
 
-You can obtain a certificate for your .DAY domain through DNSimple. To learn more, visit [SSL Certificates](/articles/ssl-certificates).
+You can [obtain a certificate for your .DAY domain](https://dnsimple.com/ssl-certificates) through DNSimple. To learn more, visit [SSL Certificates](/articles/ssl-certificates).
 
 .DAY domain names that do not have an SSL certificate configured may still be used for email and other non-browser based uses.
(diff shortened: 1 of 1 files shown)


SarahCaminiti1 (commented):
Looks good - only question - is it standard practice to swap between .DAY and .day? Is there a reason we don't stick to .DAY throughout?

OleMchls:
> Looks good - only question - is it standard practice to swap between .DAY and .day? Is there a reason we don't stick to .DAY throughout?

@SarahCaminiti1 I just copied .DE 🤷  I've a personal preference for .TLD but don't mind either way in the end. What do you prefer?

weppos:
> > Looks good - only question - is it standard practice to swap between .DAY and .day? Is there a reason we don't stick to .DAY throughout?
> 
> @SarahCaminiti1 I just copied .DE 🤷 I've a personal preference for .TLD but don't mind either way in the end. What do you prefer?

I second it. I'd prefer we stick with the UPPERCASE version consistently when referring to a specific .TLD.

content/articles/domains-day.markdown
```diff
@@ -0,0 +1,24 @@
...
+---
+
+This article explains the requirements and special procedures for .DAY domain names.
+
+
+## Registering a .DAY domain {#registering}
+
+.day is a secured namespace, meaning that websites using a .day domain require an SSL certificate to work; .day websites can only be accessed by web browsers using HTTPS through an encrypted and secured connection.
+
+You can obtain a certificate for your .day domain through DNSimple. To learn more, visit [SSL Certificates](/articles/ssl-certificates).
```
weppos, line 22: What about a call to action to the site.

```suggestion
You can [obtain a certificate for your .day domain](https://dnsimple.com/ssl-certificates) through DNSimple. To learn more, visit [SSL Certificates](/articles/ssl-certificates).
```

OleMchls: Done in [4217023](https://github.com/dnsimple/dnsimple-support/pull/949/commits/42170232600f7dd4f3055c089b35c01e8fd1ffc4)

PR #950, opened 2022-10-05
Domain list UI updates

lokst:
Update the following support pages to reflect the latest UI improvements to domain lists:
- "Organizing and Labeling Domains"
- "Record Editor"
- "Transferring a Domain to Another DNSimple Account"


dallasread (approved):
This looks great! Thanks for doing this!

sbastn (approved):
Looks good! A minor comment on the size of the new images, in some cases, they have increased by 1000% in size. Please run them through tinypng.com or another image optimizer before committing.

We use to automatically do this, but we had to remove it because of some issues.

lokst:
> Looks good! A minor comment on the size of the new images, in some cases, they have increased by 1000% in size. Please run them through tinypng.com or another image optimizer before committing.
> 
> We use to automatically do this, but we had to remove it because of some issues.

Thanks for the tip! I will get that done 👍

lokst:
@sbastn I ran the images through tinypng.com, and got about 20-30% reduction in the file sizes. Do you think these reductions are enough: https://github.com/dnsimple/dnsimple-support/pull/950/commits/1151368ebd3b35162e340644e033300c4dd5c78f ? 🙂

sbastn:
@lokst Good enough for now. Ship and we'll adjust if necessary.

lokst:
@sbastn Got it, thanks for taking a look! 🙂

PR #962, opened 2022-11-08
Promote record notes

sbastn:
I switched the Webauthn article for record notes and rotated the articles. The Netlify connector made out of the list.

# Before
Screenshot 2022-11-08 at 10 22 33

# After
Screenshot 2022-11-08 at 10 18 48


laetitiaeden (approved):
As per our discussion to promote record notes

PR #969, opened 2022-11-29
Add instructions on how to generate SSL Certificates using RSA keys

nestorsalceda:


jacegu (changes requested):
👏 great job @nestorsalceda 

I left a few suggestions. The important change to tackle is that customers should not buy a new certificate, but request the reissue of the one they just got.

content/articles/i-got-an-ecc-certificate-but-i-want-a-rsa-one.markdown
```diff
@@ -0,0 +1,57 @@
...
+
+* C is for Country
+* ST is for State
+* L is for Location
+* O is for Organization
+* CN is for Common Name, this is specially important as it must match the certificate one.
+</info>
+
+```bash
+$ openssl req -new -key KEY_FILE.pem -nodes -out CSR_FILE.pem -subj "/C=US/ST=FL/L=Melburne/O=dnsimple/CN=sep0901.jacegu.work"
```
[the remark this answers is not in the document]
nestorsalceda: Hahah, no more jacegu.work in support docs?? 🤣

content/articles/i-got-an-ecc-certificate-but-i-want-a-rsa-one.markdown
```diff
@@ -0,0 +1,57 @@
...
+```
+
+## Follow the regular process for ordering a Standard SSL Certificate
```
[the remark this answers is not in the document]
nestorsalceda: Updated the changes, thank you for the feedback! :)

categories/ssl_certificates.yaml
```diff
@@ -26,6 +26,7 @@ Ordering:
 - "Renewing a standard SSL Certificate"
 - "How can I select a different SSL certificate domain validation email?"
 - "Choosing the SSL Certificate Names"
+- "I got an ECC-signed certificate but I want a RSA one"
```
itsalyse, line 29: This should read
"I got an ECC-signed certificate, but I want an RSA one"

itsalyse: or "I got an ECC-signed certificate, but want RSA"

itsalyse:
Just a couple notes since I've noticed this in a few PRs: 

* we should use standard American english spelling since that's what's used in all the other documentation - e.g. `color` instead of `colour`

* Acronyms that begin with a consonant don't always take 'a', because indefinite articles are based on pronunciation, not letters. RSA starts with a vowel *sound*, as does SSL, so they're paired with `an` not  `a`.  

In this file, the example should say 

> I received an ECC signed certificate, and I need an RSA based one 

image

The notes here should also read:

> For instance: _I lost the private key_ or _The key has been compromised_.

and 

> You will receive the issued certificate at the email specified above.

itsalyse (changes requested):
edits needed for the image file - in separate comment 

otherwise lgtm

categories/ssl_certificates.yaml
nestorsalceda: Updated!

Thank you @itsalyse !

PR #979, opened 2022-12-23
DNSSEC: Add clarity to each of the scenarios

dallasread:
A [customer](https://dnsimple.groovehq.com/tickets/165911) was confused about how to enable DNSSEC in their specific case. This adds specific steps for each scenario to make it a little easier to follow (and makes each one linkable!).


lokst (approved):
LGTM. Great work clarifying the different scenarios a customer might use DNSSEC in 👏

itsalyse (approved):

PR #1013, opened 2023-03-15
Adds Okta as a new feature

sbastn:
This PR announces Okta as the main feature. We also rotate the features under the main one to be "record notes" and "edge network". The copy used comes directly from the Wiki page for the release of the Okta feature and the blog post titles.

Screenshot 2023-03-15 at 11 40 32


laetitiaeden (approved):
After talking to @sbastn we agree on the articles that should be showcase.

PR #1044, opened 2023-05-29
Update Okta support article

lokst:
Refresh screenshots and add minor updates and improvements


kojdm (approved):

dallasread (approved):
Thanks for doing this. Just to clarify – does it matter if users are assigned to the app BEFORE linking in DNSimple?

itsalyse (approved):

PR #1070, opened 2023-07-28
Update setting-name-servers.markdown

kojdm:
Adds additional step to navigate to new `Registration` tab when changing domain delegation


dallasread (approved):

lokst (approved):
Thanks for the update!

PR #1076, opened 2023-08-21
Google Domain Transfer Chrome Extension support article

DXTimer:
This is the support article for the Chrome extension. It covers all the steps required to get started including the initial setup of the extension for all known out-of-the-box supported browsers those being Google Chrome, Microsoft Edge, and Brave. There may be more browsers like Arc or SigmaOS but I've not had the chance to verify as they are not mainstream. Firefox is not on the list because it does not have an easy-to-follow instruction set for loading the extension from ZIP (Chrome App Store is supported but we are not listed there yet).

I have purposefully avoided giving too detailed of steps on how to "use" the extension as the extension is meant to guide the user through their transfer journey for each domain. I have only covered how to load the domains into the extension.

Belongs to https://github.com/dnsimple/dnsimple-marketing/issues/623

I have proposed to host the ZIP file at Google Drive, let me know if you prefer to have it hosted somewhere else.


itsalyse:
I've updated the article to spell out all the steps for each browser. If we're spelling out specific use cases, we should never make users scroll back and forth between different sections to find what they need.

itsalyse (approved):

laetitiaeden (approved):

PR #1083, opened 2023-09-22
Document scoped access tokens

lokst:
Document scoped access tokens


dallasread (approved):
This looks great!

What do you think about including something about seeing the last updated timestamp on the index page?

lokst:
> What do you think about including something about seeing the last updated timestamp on the index page?
@dallasread Great idea! Added 👍

kojdm (approved):
LGTM! 💯

itsalyse:
One small change to this screen: 


There's a comma splice in the instructions. It should either be 

> Keep a copy of this token — we won't show it again. 

or 

> Keep a copy of this token. We won't show it again. 


Everything else looks good!

lokst:
Thanks for catching that @itsalyse ! Just updated the screenshot 🙂

itsalyse (approved):

PR #1085, opened 2023-09-22
Refresh article about DNS query limits and fees

ggalmazor:
In this PR:
- Adapt `articles/dns-query-limits` to explain how our query volume fees work

scrnli_9_22_2023_1-47-29 PM

### DIFF
--- a/content/articles/dns-query-limits.markdown
+++ b/content/articles/dns-query-limits.markdown
@@ -1,13 +1,68 @@
 ---
-title: Do You Have Query Limits or Charge Based on Query Volume?
-excerpt: We don't implement query limits or charge based on query volume.
+title: DNS query volume fees
+excerpt: We charge $0.10 per million queries per zone.
 categories:
 - DNSimple
-- DNS
 ---
 
-# Do You Have Query Limits or Charge Based on Query Volume?
+# DNS query volume fees
 
-We don't implement query limits or charge based on query volume.
+TLDR: We charge $0.10 per million queries per zone on Solo and Teams plans.
 
-We might implement query-based pricing in the future. We'll give you plenty of advance notice if we do.
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+## How we calculate fees
+
+- The price for 1 million queries per zone is $0.10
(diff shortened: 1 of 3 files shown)


nestorsalceda (approved):
Thank you!

content/articles/dns-query-limits.markdown
```diff
@@ -1,13 +1,68 @@
...
+{:toc}
+
+---
+
+## How we calculate fees
+
+- The price for 1 million queries per zone is $0.10
+- The DNS query volume fees are calculated summing this price every time the registered volume goes beyond an x million queries mark (starting at 0 queries):
+    - The fee for 1 query is $0.10
+    - The fee for 1M queries is $0.20
```
itsalyse, line 24: If the price for one million queries per zone ins $0.10, and then we list the price here for 1M queries as $0.20, people are going to be confused. 

Would it work to reword it to "The price for up to one million queries per zone is $0.10." and then "The fee over 1M but fewer than 2M queries is $0.20"?

ggalmazor: What if we choose different examples like:

300,000 queries > $0.10
1,300,000 queries > $0.20
2,300,000 queries > $0.30

ggalmazor: Another idea:

From 1 to 999,999 queries > $0.10
From 1,000,000 to 1,999,999 queries > $0.20


Etc

ggalmazor: My intention was to explain it with specific examples on top of the abstract definition to make it more clear, but I guess it can confuse customers more…

itsalyse: I think the second one would clarify it the best — 

The price for up to, but not including, one million queries per zone is $0.10.
- The DNS query volume fees are calculated at this price every time the registered volume goes beyond any given million-query mark (starting at 0 queries):
    - The fee for 1 — 999,999 queries is $0.10
    - The fee for 1M — 1,999,999 queries is $0.20

ggalmazor: I like it! Thanks @itsalyse :)

itsalyse (approved):

PR #1096, opened 2023-10-13
Okta: Mention why we require an Okta administrator

dallasread:
When helping a customer, it took a while to track down the reason why we require an Okta administrator to do the initial set up. Let's be up-front and clear about it:

> Only super admins can view and configure event hooks.

https://help.okta.com/en-us/content/topics/automation-hooks/add-event-hooks.htm

Fixes https://github.com/dnsimple/dnsimple-support/issues/1097


dallasread:
@itsalyse can you words this better?

lokst (approved):
Thanks for documenting it!

kojdm (approved):
💯

itsalyse (approved):

PR #1130, opened 2024-02-09
Fix CAA docs regarding `issue` and `issuewild` tags

ggalmazor:
A customer provided some feedback about the wording in the docs. While I was tweaking it @weppos and I realized that some parts were actually wrong, so this PR fixes that.


content/articles/caa-record.markdown
```diff
@@ -65,9 +69,14 @@ example.com.  CAA 0 issue "letsencrypt.org"
...
-The presence of `issuewild` overrides any CAA record with the `issue` tag in the same name.
-</info>
+**Complex permissions for Let's Encrypt**
+
+```
+example.com.  CAA 0 issue "letsencrypt.org;validationmethods=dns-01"
+example.com.  CAA 0 issuewild "letsencrypt.org"
+```
+
+With this setup, Let'sEncrypt can issue certificates for example.com using the DNS validation method, while allowing any validation method for wildcard certificates only.
```
weppos, line 79:
```suggestion
With this setup, Let's Encrypt can issue certificates for example.com using the DNS validation method, while allowing any validation method for wildcard certificates only.
```

content/articles/caa-record.markdown
```diff
@@ -35,8 +35,12 @@ See the [validations](#validation) section for more details about accepted value
 ## Policies restricting certificate issuance to specific types and CAs
 
 CAA records can control the issuance of single-name certificates, wildcard certificates, or both, though the `issue` and `issuewild` tags:
-- `issue` sets a policy for single-name certificate issuance.
-- `issuewild` sets a policy for wildcard and single-name certificate issuance and overrides any `issue` policy set on the same name.
+- `issue` sets a policy for for single-name **and** wildcard certificate issuance.
+- `issuewild` sets a policy for wildcard certificate issuance.
+
+<warning>
+The presence of `issuewild` overrides any CAA record with the `issue` tag in the same name for the same CA.
```
weppos, line 42:
```suggestion
The presence of `issuewild` overrides any CAA record with the `issue` tag for the same DNS qualified name.
```

That's actually irrelevant if it's the same CA or not. If an `issuewild` it is present, it takes precedence over any `issue` for wildcard.

```
example.com.  CAA 0 issue "ca1.example"
example.com.  CAA 0 issuewild "ca2.example"
```

The example above will allow ONLY ca1 to issue single-name, and ONLY ca2 to issue wildcard.

ggalmazor: What about:

> When requesting a certificate for a wildcard domain name, the presence of `issuewild` overrides any CAA record with the `issue` tag in the same name

ggalmazor:
@weppos I've tweaked the changes based on our private conversation. Let me know what you think 🙇

itsalyse (approved):

PR #1131, opened 2024-02-12
Add embedded video

aeden:
Adds an embedded how-to video.


sbastn (approved):
Looks good. Verified on https://deploy-preview-1131--dnsimple-support.netlify.app/articles/registering-domain/

Made some minor updates so we don't hardcode the video size and it can take as much space as the rest of the content.

laetitiaeden (approved):
That looks really cool, I am very happy we're able to have video on support doc. we see how it has an impact with the number of view on youtube.

PR #1132, opened 2024-02-20
Switch mardown extension from .markdown to .md

weppos:
We had a mix of both. Tests were expecting (and checking) only `.markdown`.

I have updated the extension, updated the tests, and added a test to ensure we won't create more `.markdown` files.


sbastn (commented):
Thanks!

PR #1133, opened 2024-02-20
Add dependabot instructions to update JS files

weppos:
This config was missing, and we never updated JS dependencies. As a result, when we had a vulnerability alert recently, we dependabot was unable to open a PR to fix it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,12 +10,29 @@ updates:
       - task
       - dependencies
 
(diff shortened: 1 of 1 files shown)


dperrymorrow (approved):

weppos:
Merged. @dperrymorrow @sbastn note it opened a bunch of new PRs (as expected) which I'd like to see reviewed in the upcoming static sites maintenance iteration.

PR #1154, opened 2024-03-05
Adding meta field to frontmatter of the articles

dperrymorrow:
Duplicating the excerpt for the meta so it can be updated without affecting the display of the articles.
This will allow us to make changes for SEO purposes without affecting the display of the articles. Previously the "excerpt" was used for both the short representation of the article and the Meta.

Read below ticket for more context.
https://github.com/dnsimple/dnsimple-marketing/issues/561

### DIFF
--- a/content/articles/a-record.md
+++ b/content/articles/a-record.md
@@ -1,4 +1,5 @@
 ---
+meta: An A record maps a domain name to the IP address (IPv4) of the computer hosting the domain.
 title: What's an A Record?
 excerpt: An A record maps a domain name to the IP address (IPv4) of the computer hosting the domain.
 categories:
--- a/content/articles/aaaa-record.md
+++ b/content/articles/aaaa-record.md
@@ -1,4 +1,5 @@
 ---
+meta: An AAAA record maps a domain name to the IP address (IPv6) of the computer hosting the domain.
 title: What's an AAAA record?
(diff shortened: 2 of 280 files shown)


sbastn (approved):
👍 from my side

weppos:
@dperrymorrow @sbastn I just came across this change today. It's a lot of duplication, and also quite a bit of effort to maintain over time.

The template has been correctly designed to use the meta value first if present, and fallback to excerpt if not. Any reason then for not having the meta attribute ONLY in those articles we need to make such call? It will prevent from having tons of duplicated sentences to maintain, for which it is very hard to determine any unintentional changes.

sbastn:
@weppos The expectations at that time were different, as we anticipated a large number of articles would be updated. Currently, the majority (277 to be exact) of our article excerpts exceed the recommended length of 160 characters for SEO, with only 3 articles below this recommended length.

I checked, and as of now, none of the meta descriptions have been updated, resulting in many duplications, I agree. However, we should check with @laetitiaeden to understand the timeline for the next push on this project.

weppos:
> Currently, the majority (277 to be exact) of our article excerpts exceed the recommended length of 160 characters for SEO, with only 3 articles below this recommended length.

I suspect for many of them, a 160 char excerpt would likely be totally fine. Resulting in no need to have 2 separate fields.

PR #1172, opened 2024-05-08
Update outbound AXFR IPs

DXTimer:
We are migrating our AXFR servers to new infrastructure and these are the new IPs we want our customers to use and rely on going forward.


content/articles/secondary-dns.md
```diff
@@ -124,8 +124,8 @@ If you don't want to use Secondary DNS via AXFR, click the <label>Delete Seconda
 To set up your secondary provider, give them the server that will respond to AXFR queries (also known as DNS zone transfer). Some providers will ask for a host name, and others will ask for an IP address.
 
 The host name for the DNSimple AXFR server is **axfr.dnsimple.com**. The IP addresses on that node are:
-- `50.31.225.92`
-- `2001:1838:4000:402::a`
```
san983, line 128: Do we want to keep the former ones noted below mentioning those were switched?

DXTimer: I don't believe that's necessary. We will be sending out emails to inform all relevant parties about the deprecation of the IPs, which should address any concerns related to the old IPs.

Issue #362, opened 2018-10-08
Ruby Sass is deprecated


weppos:
```
Ruby Sass is deprecated and will be unmaintained as of 26 March 2019.

* If you use Sass as a command-line tool, we recommend using Dart Sass, the new
  primary implementation: https://sass-lang.com/install

* If you use Sass as a plug-in for a Ruby web framework, we recommend using the
  sassc gem: https://github.com/sass/sassc-ruby#readme

* For more details, please refer to the Sass blog:
  http://sass.logdown.com/posts/7081811
```

sbastn:
Nanoc does not yet support Sassc: nanoc/nanoc#1343. It's currently a feature request: https://github.com/nanoc/features/issues/45

kayla-broussard:
@sbastn @weppos while this does not resolve the issue (and is not a nanoc official solution), it appears as though someone has created a filter to assist with this since this issue was created: 

https://gist.github.com/jarrodldavis/72d450c636ffed67a0a8a3c31d261387

Nanoc themselves do not appear to be working on this. :/

PR #1174, opened 2024-05-09
Removes sass in favor of dart_sass

sbastn:
Closes https://github.com/dnsimple/dnsimple-support/issues/362

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,6 @@ gem 'kramdown-parser-gfm'
 gem 'mime-types'
 gem 'nokogiri'
 gem 'rouge'
-gem 'sass'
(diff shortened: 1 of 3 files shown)


dperrymorrow (approved):

dallasread (approved):

dallasread:
I'm getting an error when `rake run`ning:

```
soft_reference.rb:29:in `synchronize': can't be called from trap context (ThreadError)
``

Is that something you're seeing?

(I'm getting this on `main` as well, so I don't believe it is specific to this branch)

sbastn:
@dallasread I'm getting the same error on main as well.

PR #1180, opened 2024-05-22
Adds description for onboarding DNS records to a zone

sbastn:
This PR adds a section on the DNS hosting page describing the new onboarding DNS page.

See https://deploy-preview-1180--dnsimple-support.netlify.app/articles/dns-hosting/#onboarding-dns-records-to-your-domain for preview

### DIFF
--- a/content/articles/dns-hosting.md
+++ b/content/articles/dns-hosting.md
@@ -39,12 +39,12 @@ To use our DNS hosting service, the account must be subscribed to a valid plan. 
 
 To use our DNS hosting services, you need to [add the domain to a DNSimple account](/articles/adding-domain). If you also want to use DNSimple as registrar, you may need to register or [transfer](/articles/transferring-domain) the domain to us.
 
-## Onboarding DNS records to your domain
+## Adding DNS records to your domain
 
 Once you've added your domain, you can configure the DNS records using the following options:
 
 - By [auto-importing](https://support.dnsimple.com/articles/auto-import-dns/) existing DNS records if your zone is resolving with another DNS provider.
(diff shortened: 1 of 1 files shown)


sbastn:
@itsalyse can I please have your review on this PR? Thank you!

stephanoodle:
> @Pemacaa or @stephanoodle could one of you take a look at this too? This article will have further updates later, but want to make sure anything we're adding in the meantime is reviewed.

I took a look. Here's what I found:

> Onboarding DNS records

Why was onboarding chosen vs adding? We typically tell customers to add records. 

> Configuring [two-way connectors](https://support.dnsimple.com/categories/connectors/). With two-way connections, both sides of the service are configured and monitored automatically.

We do not mention anywhere that I am aware of in the support documentation that Heroku and Netlify are two-way connectors. This is a new term and if we are using it, it should probably be mentioned in those two docs at least. 

> Manually entering DNS records in the DNS [record editor](https://support.dnsimple.com/articles/record-editor/).

Right below this, we mention the DNS record editor, however, the link is for the entirety of [DNS record editor]. 

Other than that, this looks good to me.

Pemacaa:
I have the same question as @stephanoodle about why we are using "Onboarding DNS records" instead of adding. The rest looks good to me.

itsalyse (approved):

sbastn:
Thanks @Pemacaa and @stephanoodle for your feedback. I've incorporated it to the changes.

I used 'two-way connectors' since it's how we call them here: https://dnsimple.com/benefits/integrations but you're right that we don't use the same concept in the support docs, so I replaced it.

PR #1189, opened 2024-06-06
Updates to changing email article

dallasread:
Belongs to https://github.com/dnsimple/dnsimple-business/issues/1975


content/articles/changing-email.md
```diff
@@ -15,16 +15,6 @@ categories:
 
 ---
 
-## A note about accounts
```
dallasread, line 18: I think we can remove all of this if we agree to this: https://github.com/dnsimple/dnsimple-app/pull/27073/files#r1629893283

jacegu: 👍 Agreed.

jacegu:
Reading through the article after the changes, I realized that in the _Changing the user email_ section, we have the following:

> 1. If the update is successful, you'll be redirected to the account page. Make sure the account correctly displays the new address.

This is not true, it should say _redirected to the **user** page_ instead. We should take the opportunity to update this.

jacegu:
You are going to hate me for this, but we should also take the opportunity and update the screenshots for the update account email page, as we now have all the information about the profile in there as well.

jacegu (approved):

dallasread:
🙌 Thanks you!

content/articles/changing-email.md
```diff
@@ -15,16 +15,6 @@ categories:
...
-
-When you initially sign up as a user in our system, we create an account bound to that user. We call this the Primary Account, and the email address is the same as the username. If you change the email address for that user, the account email address will also be updated. That way customers with a single account only have to update their email in one place. Follow the instructions in [Changing User Email](#changing-the-user-email) if this is the case for you.
-
-Extra accounts have their own email addresses that can be changed ad-hoc.
-
-<warning>
-System limitations do not currently allow Primary Accounts to be deleted.
-</warning>
-
 ## Changing the account email
```
stephanoodle, line 18: I'm confused here and think I am missing something. 

I thought we were changing this so that you only have a user email. Why would there be an account email as well?

stephanoodle: Never mind. I just looked at https://github.com/dnsimple/dnsimple-app/pull/27073/files and see what you're doing. The two are no longer connected so updating one does not update the other.

stephanoodle (approved):

content/articles/changing-email.md
jacegu: That's exactly it. Thanks for taking a look @stephanoodle ❤️

itsalyse (approved):

content/articles/changing-email.md
dallasread: Perhaps @itsalyse can a find a place to clarify this: 🥺 
1. Account email is the one that receives notifications (eg. expiring domains, billing, etc).
2. User email is just for logging in (and user-specific notifications like SSO, reset password).

PR #1225, opened 2024-07-31
Update DNSSEC & SecondaryDNS RFC

onlyhavecans:
A customer wrote in about the technical accuracy of this article. While reviewing it, I noticed the RFC was finalized in 2020. 
The first purpose of this PR is to fix that.

Offline record signing was mentioned as another solution to address this, but we do not support that at this time, and I am not fully able to vouch for whether signed non-existence is fully solved for pre-signed AXFR. We could either note this or look into this as a possibility but before that...

While reviewing this article, should we rephrase or replace it as it applies explicitly to DNSimple? This reasonably technical document explains potential issues around something we don't support. I wonder if explaining "the why" is as valuable to our customers. It could become more about clarifying what we support and what customers can do depending on their use case.
Before I refactored, I wanted to bring up the conversation.

I'm CCing @m0rcq on this as a second pair of high-technical eyes.

### DIFF
--- a/content/articles/dnssec-and-secondary-dns.md
+++ b/content/articles/dnssec-and-secondary-dns.md
@@ -7,8 +7,11 @@ categories:
 
 # Why DNSSEC and Secondary DNS may not work together
 
+> [!NOTE]
+> This pertains to online signing of RR sets. DNSimple does not support offline signing of RR sets at this time.
+
 All authoritative name servers MUST sign all record sets with all private keys that are Zone Signing Key (ZSK) type. This is necessary because a resolver could get the ZSK from one authoritative name server and the RRSIG from another authoritative name server. To include the RRSIG for all ZSK DNSKEYs, the authoritative name servers must have all private key material. We currently do not share private key material, and even if we did, AXFR does not support transferring private key material and thus key rotations would have to be done manually in a coordinated fashion.
 
 Note that in practice, it is possible to run multi-provider DNSSEC without sharing ZSK private key material, however it is not guaranteed to work due to the condition described above with resolvers getting the DNSKEY and the RRset + RRSIG from different authoritative name servers.
 
-You can read more information about multi-provider DNSSEC in this [Draft RFC here](https://tools.ietf.org/html/draft-ietf-dnsop-multi-provider-dnssec-05).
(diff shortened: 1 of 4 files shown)


stephanoodle:
NGL, the Groove conversation was too technical for me to understand to verify anything with it. 

What I will say is, that linking the correct RFC in the documentation is a good move. The documentation should also be clear on what we do and do not support at DNSimple.

itsalyse:
Do we have numbers for how many customers are actually clicking through to/finding this article?

onlyhavecans:
> Do we have numbers for how many customers are actually clicking through to/finding this article? 

I do not have those numbers; maybe @sbastn can offer better information on the metrics.

I went ahead and updated this PR to reflect your suggestion. It seems to flow better and makes a clearer experience with fewer clicks for those who are not highly technical.

While I was there, I fixed some glaring typos in my eyeline.

sbastn:
> Do we have numbers for how many customers are actually clicking through to/finding this article?

Year to date, here are the numbers for this page compared to other DNSSEC and Secondary DNS pages:

- [The article](https://support.dnsimple.com/articles/dnssec-and-secondary-dns/) we're reviewing in this PR received 179 visits.
- The [DNSSEC](https://support.dnsimple.com/articles/dnssec/) article received 864 visits.
- The [Add a secondary DNS server to DNSimple](https://support.dnsimple.com/articles/secondary-dns/) article received 559 visits.

Traffic sources for this document include:
- Google: 101 referrals
- Direct: 39 referrals
- The [DNSSEC article ](https://support.dnsimple.com/articles/dnssec/): 8 referrals
- [DNSimple-app](https://dnsimple.com): 4 referrals
- The The [Add a secondary DNS server to DNSimple](https://support.dnsimple.com/articles/secondary-dns/): 4 referrals

>  I'm wondering if it might be easier to remove this doc entirely and just bake it into the warning that's on the secondary and DNSSEC docs

Sounds good to me. Right now we have the following alerts at the top of these docs:

Screenshot 2024-08-02 at 08 41 57

stephanoodle (approved):

Pemacaa (approved):

itsalyse (approved):

PR #1246, opened 2024-09-11
Update DNSMadeEasy axfr address and screenshots

AGS4NO:
- Update DNSimple outbound AXFR primary server address
- Update screenshots


sbastn (approved):
Just so you're aware, we'll have to do this as well soon: https://github.com/dnsimple/dnsimple-app/issues/28053

PR #1268, opened 2024-10-07
Add redirect to AU article

kntsoriano:
As raised [here](https://dnsimple.slack.com/archives/C07ML213JE6/p1728335640345409), we sent out the wrong article link to our AU customers. This is a PR implementing a redirect to the correct article link.


kntsoriano:
Checked with deploy preview.

PR #1274, opened 2024-10-22
making the header sticky

dperrymorrow:
https://github.com/dnsimple/dnsimple-marketing/issues/746

Make the global header sticky to promote use of search, and keeping the signup / contact sales in the header always in view.

### DIFF
--- a/content/assets/css/_application.scss
+++ b/content/assets/css/_application.scss
@@ -76,8 +76,6 @@
   }
 }
 
-
-
 // Article
 // -------------------------
 // Styles for a single article entry.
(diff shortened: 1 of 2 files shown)


dperrymorrow:
> Works great, but why not using Tachyons classes? Especially since they're only needed in the header?

https://tachyons.io/docs/layout/position/ tachyons does not include `.sticky` class for positioning. Tachyons dates back to when sticky did not have wide support

> css also provides an option for setting position to 'sticky' but it is not included in this module. If you do find it useful you could easily extend the position module to include this option. It should be noted that sticky is not well supported at the moment. If the provided coordinates don’t suit your needs the module is easy to extend to add any values you might need.

PR #1307, opened 2024-12-02
Capture search event

sbastn:
To better understand how to enhance the search functionality for an upcoming project, I'd like to track the specific terms and queries that our users are searching for. This will help us identify patterns, prioritize improvements, and ensure the search results are more relevant to their needs.

I've created a `debounce` method so we don't track every single search character. This means that that the search is behaving a bit differently. We wait 300ms after user input to submit the query.

The queries are sent to PostHog.

## QA

There is no PostHog environment configured for local testing, so your browser console will show you an error. If you want to verify the values sent to PostHog, override the `trackSearch` method to only output the query to the console:
```js
  var trackSearch = function trackSearch (query) {
    // window.posthog.capture('support-search', { query: $input.value });
    console.log(query);
  };
```

### DIFF
--- a/content/search/results.js
+++ b/content/search/results.js
@@ -52,14 +52,10 @@
   };
 
   var debounce = function debounce (func, delay) {
-    var timeout;
-    return function () {
-      var context = this;
-      var args = arguments;
+    let timeout;
+    return function(...args) {
       clearTimeout(timeout);
-      timeout = setTimeout(function () {
-        func.call(context, args[0]);
-      }, delay);
+      timeout = setTimeout(() => func.apply(this, args), delay);
     };
   };
 
--- a/content/search/results.js
+++ b/content/search/results.js
@@ -51,6 +51,18 @@
     return $results;
   };
 
+  var debounce = function debounce (func, delay) {
+    var timeout;
+    return function () {
+      var context = this;
+      var args = arguments;
+      clearTimeout(timeout);
+      timeout = setTimeout(function () {
+        func.call(context, args[0]);
+      }, delay);
+    };
+  };
+
   if (typeof module === 'object' && typeof module.exports === 'object')
     module.exports = {
       parseQueryParams: parseQueryParams,
@@ -65,8 +77,8 @@
     if (q.length)
       showResults($main, $input, q);
 
-    $input.oninput = function (e) {
+    $input.addEventListener('input', debounce(function () {
       showResults($main, $input, $input.value);
-    };
+    }, 300));
   }
 })();
--- a/content/search/results.js
+++ b/content/search/results.js
@@ -63,6 +63,10 @@
     };
   };
 
+  var trackSearch = function trackSearch (query) {
+    window.posthog.capture('support-search', { query: $input.value });
+  };
+
   if (typeof module === 'object' && typeof module.exports === 'object')
     module.exports = {
       parseQueryParams: parseQueryParams,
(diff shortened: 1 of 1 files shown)


dallasread (commented):
What happens when `window.posthog` doesn't exist?

image

content/search/results.js
```diff
@@ -51,6 +51,18 @@
...
+  var debounce = function debounce (func, delay) {
+    let timeout;
+    return function (...args) {
+      clearTimeout(timeout);
+      timeout = setTimeout(() => func.apply(this, args), delay);
+    };
+  };
+
+  var trackSearch = function trackSearch (query) {
+    window.posthog.capture('support-search', { query: $input.value });
```
dperrymorrow, line 63:
```suggestion
    window.posthog ||= { capture: console.log };
    window.posthog.capture('support-search', { query: $input.value });
```

dperrymorrow: you could define posthog if non-existant and then you would get logging without an error locally

content/search/results.js
```diff
@@ -62,11 +74,14 @@
...
+    if (q.length) {
       showResults($main, $input, q);
+      trackSearch(q);
+    }
 
-    $input.oninput = function (e) {
+    $input.addEventListener('input', debounce(function () {
       showResults($main, $input, $input.value);
-    };
+      trackSearch($input.value);
```
dperrymorrow, line 84: did you intentionally debounce the showing of the search results, or did you just want to debounce the tracking?

dperrymorrow (approved):
personally id like to see it not throw errors in local dev, as someone who comes along might think that something needs fixed. But I understand you might be just gathering data for a bit.

content/search/results.js
sbastn: @dperrymorrow originally I wanted to only debounce the tracking, but then I would need 2 listeners. Would that be OK?

content/search/results.js
sbastn: I'll create something similar to what we have in dnsimple-app

sbastn: ✅ Done

sbastn:
@dallasread @dperrymorrow I made some improvements based on your suggestions:
1. We now have a dummy PostHog client, so we don't throw an error on local
2. I've restored the original search behavior, by debouncing only the tracking of the queries

I'm requesting another review if you don't mind.

layouts/analytics.html
```diff
@@ -10,4 +10,13 @@
...
   })
 </script>
+<% else %>
+<script>
+  // Dummy PostHog client
+  const posthog = {};
+  posthog.init = function () {};
+  posthog.identify = function (arg) { console.log(`PostHog identify: ${arg}`); };
+  posthog.capture = function (event, data) { console.log(`PostHog capture - Event: ${event} - Data: ${JSON.stringify(data)}`); };
+  window.posthog = posthog;
```
dperrymorrow, line 20:
```suggestion
  // Dummy PostHog client
  window.posthog = {
    init() {},
    identify(arg) { 
      console.log(`PostHog identify: ${arg}`); 
    },
    capture(event, data) { 
      console.log(`PostHog capture - Event: ${event}`, data);
    }
  }
```
personal preference, but would terser to define the dummy object at once, also IMO its more useful to see the data as an object _(second param passed to console.log)_ rather than a JSON string

dperrymorrow (approved):

dallasread (approved):

PR #1330, opened 2024-12-19
Bring the support widget to the support site

dallasread:
Load up a support widget on an empty page in dev mode with hot-reloading, etc:

```
yarn widget
```

You can add the support widget on _any website_ (with a permitted csp policy like dnsimple.com or blog.dnsimple.com):

```js

const $div = document.createElement('div')
$div.id = 'support-widget'
document.body.appendChild($div);

$script.src = `https://deploy-preview-1330--dnsimple-support.netlify.app/widget.js`;

```


kojdm (approved):
LGTM!

PR #1331, opened 2024-12-19
Trigger to open the support widget

dallasread:
This PR makes the support widget fill more of the screen, as well:

- Adds `data-dnsimple-open-support-widget`: can be added to any DOM element to trigger the opening of the widget.
- Adds `data-dnsimple-current-site-url="https://NOTHING.dnsimple.com"`: used to determine if the article is for the site the customer is already on. For example, if the customer is @ dnsimple.com and they search for alias, we want the result to show in the support widget browser. If they are on support.dnsimple.com, then we want to do a full page load to the result.

image

## QA

Test using `yarn widget`:

- Search for `alias`.
- The search result links should open in the widget (because `data-dnsimple-current-site-url="https://NOTHING.dnsimple.com"` doesn't match the current site).

Test the [preview](https://deploy-preview-1331--dnsimple-support.netlify.app/):

- Search for `alias`.
- Click on a result should change the page like a normal page click (not use the support widget browser OR be in a new tab).

Test on DNSimple.com:

```js

$script.src = `https://deploy-preview-1331--dnsimple-support.netlify.app/widget.js`;

```

- The search result links should open in the widget.


dallasread:
I think we should make same-site links open normally (eg. without new tab, not in support widget) before releasing this.

_widget/src/components/app/component.vue
```diff
@@ -101,27 +114,33 @@ export default {
     },
 
     focus () {
-      const $header = this.$refs.header;
+      nextTick(() => {
```
dallasread, line 117: Need to wait a tick before the input is rendered.

_widget/src/components/app/component.vue
```diff
@@ -101,27 +114,33 @@ export default {
...
     fetchArticles (done) {
+      if (this.isFetched) return done();
+
       const script = document.createElement('script');
 
       this.isLoading = true;
 
       script.type = 'text/javascript';
-      script.src = `${this.rootURL}/search.js`;
+      script.src = `https://support.dnsimple.com/search.js`;
```
dallasread, line 137: This'll be cleaned up when we add multiple sources (eg. dev site articles).

package.json
```diff
@@ -32,8 +32,8 @@
   "scripts": {
     "live": "concurrently 'bundle exec nanoc live' 'bundle exec rake compile'",
     "build": "./node_modules/.bin/webpack && vite build ./_widget",
-    "lint": "eslint _widget content spec",
-    "lint:fix": "eslint _widget content spec --fix",
+    "lint": "eslint _widget src spec",
```
dallasread, line 35: :robot: was complaining.

templates/search.js.erb
dallasread, line 1: Can't wait to not make this an `erb` file. 😄

kojdm (approved):
Works really well! 🙌

PR #1354, opened 2025-01-17
Removing no longer referenced images in support articles

dperrymorrow:
Many images have been orphaned _(unused)_ do to the ongoing support updates, and updated screenshots. 
Using a local script I identified and removed images that were no longer reference


sbastn (approved):
Thank you, this really helps!

dperrymorrow:
@sbastn do you think its worthwhile to add the script i used to cleanup as an NPM script so that we can easily cleanup in the future?

sbastn:
@dperrymorrow good idea. We can start by adding it as a test, so if a PR removes an image that is no longer used, the CI build fails.

dperrymorrow:
> @dperrymorrow good idea. We can start by adding it as a test, so if a PR removes an image that is no longer used, the CI build fails. We already perform some tests on images, have a look at `_test/images_test.rb`

Oh I had just added an NPM script to remove unused, or would you rather fail the build? https://github.com/dnsimple/dnsimple-support/pull/1354/commits/1c50f5fd88b903702db10ecb4720f8fd024f2633

sbastn:
>  I had just added an NPM script to remove unused, or would you rather fail the build? https://github.com/dnsimple/dnsimple-support/commit/1c50f5fd88b903702db10ecb4720f8fd024f2633

@dperrymorrow I rather fail the build if you don't mind. The NPM script is great, but then we have to introduce process changes: When do we run this script? How often? Who does this? 

The advantage of failing the CI build is that we don't have to introduce changes to the SOP, it's also less reactive (and maybe less confusing to understand why an image is no longer referenced in the content) because the person doing the change will notice that they left the image behind after removing it from the content when their PR fails the CI build.

I have added the test on https://github.com/dnsimple/dnsimple-support/pull/1354/commits/0a4e9632410805fb86ac2a7f834795df3a1fc577 so you can see what I mean.

If you agree, let's keep the test and remove the npm script. Let me know if this makes sense.

dperrymorrow:
> If you agree, let's keep the test and remove the npm script. Let me know if this makes sense.

Sure, I have removed the NPM script

PR #1359, opened 2025-01-23
Bump video up on the a-record page 

dperrymorrow:
Move video above the headline so that the title of the article is not confusing with the title of the video.


laetitiaeden (approved):
Further to our chat to move the video above the question What is an A record?

sbastn (approved):
Thanks!

PR #1363, opened 2025-02-06
Adding the new video to diff between record types

dperrymorrow:
**New video:**
https://youtu.be/j8XxjL3hNYk

Adding it to the correlating page. 
`/articles/differences-a-cname-records/`

_(PS)_ also changed `overflow-x: auto` on `<pre>` tags so that it only shows scrollbars if needed in our code examples, instead of always showing them.

### DIFF
--- a/content/assets/css/_typography.scss
+++ b/content/assets/css/_typography.scss
@@ -72,7 +72,7 @@ code {
 
 pre {
   display: block;
-  overflow-x: scroll;
+  overflow-x: auto;
   padding: 10px;
   font-size: 100%;
   font-family: $font-monospace;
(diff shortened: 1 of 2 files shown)


dperrymorrow:
Screenshot 2025-02-06 at 12 15 40 PM

content/assets/css/_typography.scss
```diff
@@ -72,7 +72,7 @@ code {
 
 pre {
   display: block;
-  overflow-x: scroll;
+  overflow-x: auto;
```
sbastn, line 75: Why this change?

sbastn (approved):
Thank you!

PR #1369, opened 2025-02-11
Search: "Getting started" driven by article

dallasread:
This PR fixes a bug due to our Getting Started page linking to relative paths. This would break when embedded on DNSimple.com.

Now, we load the Getting Started support article in its place, so everything is linkified already (and means we only have 1 source of truth).

### Locally
  - `yarn && yarn widget`
  - The support widget should open with the "Getting started" article and links should work

### On dnsimple.com

```js

$script.src = `https://deploy-preview-1369--dnsimple-support.netlify.app/widget.js`;

```


sbastn (approved):
🚀

kojdm (approved):
QA ✅ Thanks!

PR #1371, opened 2025-02-12
Search: Improve results

dallasread:
This PR improves the results that are laid out in the [frame](https://docs.google.com/document/d/1CykewBhR1lRwDzA6uwmyZ_nkKhg36xuKaQvOQg0MyHQ/edit?tab=t.0) & [shape](https://miro.com/app/board/uXjVL-b6wd0=/?moveToWidget=3458764610826940461&cot=14). Also addresses https://dnsimple.slack.com/archives/C047A16BDU5/p1739382124953689?thread_ts=1739354089.937599&cid=C047A16BDU5.

## 👓 Preview & QA

https://deploy-preview-1371--dnsimple-support.netlify.app/


_widget/src/components/app/component.vue
```diff
@@ -73,7 +72,7 @@ export default {
 
   watch: {
     q (val) {
-      if (val.length > 2) {
+      if (val.length > 1) {
```
dallasread, line 75: Some record types are shorter than 2 characters. Sadly, A records are still left out.

dallasread: The reason for filtering these out is that a query of a single character is very difficult to determine intent. Something like `mx` has a clear intent.

spec/_widget/components/search.spec.js
```diff
@@ -1,4 +1,5 @@
-import { search, prepareArticles, dictionaryTermMatches, articleScore, fixRelativeImgSrcs } from '../../../_widget/src/components/app/search.js';
+import Search from '../../../_widget/src/components/app/search.js';
+import ARTICLES from '../../../output/search.json';
```
dallasread, line 2: There's a trade-off here in that we need to compile the static site before running the specs... but we get great data to work with.

_widget/src/components/app/component.vue
```diff
@@ -52,17 +49,20 @@ export default {
...
         return window.fetch(url).then((r) => r.json());
       }
+    },
+    search: {
+      type: Object,
+      default() { return new Search(); }
     }
   },
   data () {
-    const query = this.query || urlMatchingDictionary(window.location.href);
```
dallasread, line 58: We weren't using `query` and I thought it was a good name for the searching method. 🤷‍♂️

_widget/src/components/app/component.vue
kojdm: Maybe we could explicitly account for record types in the value? I.e. if the query value is "A", we show the A record articles. 

Related: currently in this branch, when searching `a record`, it shows the most relevant article last. Compared to the results in current production where the most relevant article is at the top.

| deploy-preview-1371--dnsimple-support.netlify.app | support.dnsimple.com |
|--------|--------|
| image | image |

kojdm (approved):
🔥 Awesome! Just left a comment about some improvement when searching for A records, but it worked great for all the other queries I tested.

_widget/src/components/app/component.vue
dallasread: Hm... good find.

kojdm: Just tested and it's a lot better. Thanks!

PR #1375, opened 2025-02-13
Search: Include results from developer.dnsimple.com

dallasread:
This PR adds results from the dev site.

## QA

image


_widget/src/components/app/component.vue
```diff
@@ -53,6 +57,15 @@ export default {
     search: {
       type: Object,
       default() { return new Search(); }
+    },
+    sources: {
+      type: Array,
+      default() {
+        return [
+          { name: 'DNSimple Support', url: 'https://support.dnsimple.com' },
+          { name: 'DNSimple Developer', url: 'https://deploy-preview-760--dnsimple-developer.netlify.app' }
```
dallasread, line 66: Right now, it doesn't support CORS yet until https://github.com/dnsimple/dnsimple-developer/pull/760:

```suggestion
          { name: 'DNSimple Developer', url: 'https://developer.dnsimple.com' }
```

_widget/src/components/app/component.vue
```diff
@@ -44,6 +44,10 @@ export default {
       type: String,
       default: ''
     },
+    gettingStartedUrl: {
+      type: String,
+      default: '/articles/getting-started/'
```
dallasread, line 49: We're lucky there is no overlap of URLs between the sites. But this could be a potential issue in the future.

kojdm: Maybe we can modify `findArticle` to accept a source? That way it only returns an article with a matching `sourceUrl`.

```js
this.findArticle(this.gettingStartedUrl, 'https://support.dnsimple.com')
```

kojdm (approved):
QA ✅ Works great!

_widget/src/components/app/component.vue
dallasread: Changed `findArticle` to just use the absolute URL: https://github.com/dnsimple/dnsimple-support/pull/1375/commits/0bf761b1c6167595794b4e18c693ed8ce1672a04

PR #1376, opened 2025-02-13
Fix: Change article source back to support.dnsimple.com

kojdm:
Fixes a bug in the search widget where clicking on articles in support.dnsimple.com would open them in the widget, and not redirect to the actual article.

### DIFF
--- a/_widget/src/components/app/search.js
+++ b/_widget/src/components/app/search.js
@@ -58,7 +58,7 @@ const prepareArticles = (articles, source) => {
     article.searchBody = article.searchBody || (article.body || '').toLowerCase().replace(PUNCTUATION, '');
     article.body = fixRelativeImgSrcs(article.body || '');
     article.categories = article.categories || [];
-    article.source = source;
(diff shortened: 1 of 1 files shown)


kojdm:
✅ Verified that the articles have the correct URLs

image

dominicumbeer:
Does this fix work when the search widget is used on developer.dnsimple.com or the search result links to a [page that is not part of the support website](https://dnsimple.slack.com/archives/C08CHR32Z2M/p1739404763003609)?

kojdm:
> Does this fix work when the search widget is used on developer.dnsimple.com or the search result links to a [page that is not part of the support website](https://dnsimple.slack.com/archives/C08CHR32Z2M/p1739404763003609)?

@dominicumbeer No, not yet. This was only a temporary fix to restore the old functionality. New functionality like you described will be added in https://github.com/dnsimple/dnsimple-support/pull/1371 and https://github.com/dnsimple/dnsimple-support/pull/1375.

PR #1386, opened 2025-02-14
Search: Add an error state when remote data is unable to be loaded

dallasread:
This PR adds an error page to be used when articles are not loaded.

Fixes https://github.com/dnsimple/dnsimple-app/issues/29649
Belongs to https://github.com/dnsimple/dnsimple-business/issues/2112

## 👓 Preview

image

image

## QA

- Visit https://deploy-preview-1386--dnsimple-support.netlify.app/
- Block the request URL
- You should see the preview above

image

Alternatively, `yarn widget` and update the `search.json` ajax url to something bogus.


_widget/src/components/article/component.vue
```diff
@@ -1,10 +1,10 @@
 <template>
-  <div v-if="article" class="route with-footer">
+  <div class="route with-footer">
```
dallasread, line 2: Now, we should not be ever getting here without an article.

kojdm:
Testing this in dnsimple.com and it seems to break searching

```js
[...document.querySelectorAll('#dnsimple-support-widget')].forEach((el) => el && document.body.removeChild(el))

$script.src = `https://deploy-preview-1386--dnsimple-support.netlify.app/widget.js`;

```

image

dallasread:
I believe the issue here is that there is no getting started article yet on dnsimple.com. It will have one once [this](https://github.com/dnsimple/dnsimple-app/pull/29662) is merged.

dallasread:
Yup, that solved it.

PR #1391, opened 2025-02-19
Improving Search Experience: Recently visited articles

kojdm:
Stores a list of 10 recently visited article URLs in localStorage. Displays the articles in lieu of `Getting Started` if present.

> [!NOTE]
Recently visited articles are not shared between the different sites (dnsimple-app, dnsimple-support, dnsimple-developer) since localStorage does not work across different origins. We would probably have to use a cookie accessible from `.dnsimple.com` subdomains instead of localStorage for this to work.
>  

image

## QA

- Visit https://deploy-preview-1391--dnsimple-support.netlify.app/
- Do some searching
- Visit some articles
- Verify that the recently visited articles are displayed when:
  - Opening the widget (if the search query is empty)
  - Deleting the search query


dallasread (commented):
QA'd 💯. A couple thoughts:

- Seeing the recently visited articles is really nice.
- I think it's fine that the history is not cross-site. It is something we could switch to cookies at a later date anyway.
- I wonder if we should be adding to the history whenever an article is visited – both in the widget and just "vanilla" visiting the article (eg. we could track it when the support widget is loaded). The reason why I'm thinking about that is that I was clicking around using the sidebar to visit sites and they weren't appearing in the recently visited list.
- What do you think about lowering the number of stored articles to 5, but also include the getting started article below the list? Or maybe only show the article history if there are at least 3 items (it just looks so empty to me)?

dallasread (approved):
QA'd 💯.

kojdm:
> - I wonder if we should be adding to the history whenever an article is visited – both in the widget and just "vanilla" visiting the article (eg. we could track it when the support widget is loaded). The reason why I'm thinking about that is that I was clicking around using the sidebar to visit sites and they weren't appearing in the recently visited list.
> - What do you think about lowering the number of stored articles to 5, but also include the getting started article below the list? Or maybe only show the article history if there are at least 3 items (it just looks so empty to me)?

Added tickets to our board to address these concerns.

PR #1395, opened 2025-02-20
Search: Better results for partial matches

dallasread:
This improves dictionary-matching for partial words. It does have a very slight degradation for some results. The trade-off seems worth it as we're getting effective results with a very small query.

### DIFF
--- a/_widget/src/components/app/dictionary.js
+++ b/_widget/src/components/app/dictionary.js
@@ -11,7 +11,6 @@ export default {
   "letsencrypt": "lets encrypt",
   "membership": "share management",
   "team": "member",
-  "annual": "yearly",
   "remove": "delete",
   "delegate": "registrar point setting",
   "payment": "invoice payment",
--- a/_widget/src/components/app/dictionary.js
+++ b/_widget/src/components/app/dictionary.js
@@ -2,8 +2,8 @@
 // Dictionary only works with lowercase with no punctuation
 
 export default {
-  "add user": "member",
-  "remove user": "member",
+  "add user": "multiple team",
+  "remove user": "multiple team",
   "user": "user settings",
   "close": "unsubscribe",
   "cancel": "unsubscribe",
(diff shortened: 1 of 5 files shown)


dallasread:
I'm not sure this is worth merging – some results are better, some are not.

dallasread:
`expiry` => `expiration`

_widget/src/components/app/dictionary.js
```diff
@@ -2,25 +2,24 @@
...
   "remove": "delete",
   "delegate": "registrar point setting",
   "payment": "invoice payment",
   "transfer": "transfer dnsimple",
   "nameservers": "name servers",
-  "domain": "register",
   "metrics": "query volume analytics",
   "stats": "query volume analytics",
   "observability": "query volume analytics",
+  "query": "query volume analytics",
```
dallasread, line 22:
```suggestion
  "expiry": "expiration",
```

spec/_widget/components/search.spec.js
```diff
@@ -42,10 +42,24 @@ describe('Search', () => {
...
+    const results1 = subject.query('expi');
+
+    expect(results1.map((r) => r.title).slice(0, 2)).toContain('Product Expiration Notification');
+  });
+
   describe('queries', () => {
     const queries = {
       'a record': {
-        'Managing A Records': 4,
+        'Managing A Records': 6,
```
dallasread, line 62: Most of these results just moved a little bit. I think the trade-off is worth it considering we get completions for partial words.

_widget/src/components/app/dictionary.js
```diff
@@ -1,27 +1,30 @@
...
   "queries": "query volume analytics",
   "analytics": "query volume analytics",
+  "expiry": "expiration",
   "comment": "note"
```
dallasread, line 29:
```suggestion
  "comment": "note",
  "axfr": "secondary dns"
```

kojdm (approved):

PR #1400, opened 2025-02-23
Search: Refine the button

dallasread:
This PR makes our search button look and feel more like a button.

## 👓 Preview

image

image

image

image

image

image


kojdm:
There's this blue line that appears next to `dnsimple SUPPORT` in the header... Not sure if it's intentional.

image

kojdm (commented):
Looks great! Should we do the same for the developer site?

kojdm (approved):

PR #1407, opened 2025-02-26
Improve results for `domain validation`

dallasread:
This PR adds a dictionary item for `validation` so that our [ICANN Domain Validation Requirements](https://support.dnsimple.com/articles/icann-domain-validation/) article shows up.

The case that was brought up was `domain validation`. With our new dictionary entry, this will be converted to `domain icann validation`, so the "domain" part of the query is still preserved. As a general rule, I've noticed that 1-word dictionary terms that simply add additional words (like in this PR) are preferable – so that we don't skew other results (eg. results for `certificate validation`).

Related to https://dnsimple.slack.com/archives/C08CHR32Z2M/p1740563758239829?thread_ts=1740068284.101949&cid=C08CHR32Z2M

## 👓 Preview

image

### DIFF
--- a/_widget/src/components/app/dictionary.js
+++ b/_widget/src/components/app/dictionary.js
@@ -28,4 +28,5 @@ export default {
   "expiry": "expiration",
   "comment": "note",
   "axfr": "secondary dns",
+  "validation": "icann validation"
 };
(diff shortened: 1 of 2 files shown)


sbastn (approved):
Works!

Screenshot 2025-02-26 at 12 20 13

PR #1408, opened 2025-02-26
Improving Search Experience: Undo category grouping

kojdm:
Don't group search results by category. Sometimes, matching articles would be pushed down the results due to the grouping.

image

### DIFF
--- a/_widget/src/components/articles/component.vue
+++ b/_widget/src/components/articles/component.vue
@@ -27,32 +27,28 @@
     </div>
 
     <div v-else class="articles">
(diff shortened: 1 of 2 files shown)


dallasread (approved):
Qa'd fine. I'm torn on this because I like the categories. 

I wonder if we could order the categories by their cumulative articles' score so that the most relevant category floats to the top. If we keep categories, I think we should the specs around query ranking be checking what the customer actually sees (NOT simply the output of the search logic).

Whatever you think!

PR #1413, opened 2025-02-28
Delete deprecation articles

weppos:
This PR removes several references to changes we rolled out in 2024.


dallasread (approved):
Thanks!

stephanoodle (approved):

dallasread:
It looks like the search query specs were failing because we had specified for this article to be in the results when `axfr` was searched. I have removed it from the spec.

stephanoodle (approved):

PR #1419, opened 2025-03-06
Search: Improve url matching terms

dallasread:
This PR adds some items from https://docs.google.com/spreadsheets/d/1L3XtVeMh_8ADjF23WuV024rE9bhpc-GPI7xlXUGUnQI/edit?gid=240358090#gid=240358090 to the URL-matching dictionary.


_widget/src/components/app/component.vue
```diff
@@ -135,7 +136,7 @@ export default {
     },
 
     showRecentlyVisitedArticles () {
-      return this.q.length === 0 && this.recentlyVisitedArticles?.length > 0;
+      return this.q.length === 0 && this.filteredArticles.length === 0 && this.recentlyVisitedArticles?.length > 0;
```
dallasread, line 139: This was essentially never showing the URL-matched articles. Now, we show them if they exist. 

You can see your history at any time by clearing the query.

_widget/src/components/app/url-dictionary.js
```diff
@@ -7,19 +7,27 @@ const DICTIONARY = {
   'Standard SSL Certificate': /standard/,
   'Reverse Zones': /zones/,
   'share management': /memberships/,
-  Automation: /automation/,
+  'Automation': /automation/,
```
dallasread, line 10: We could use some regression specs on these, but I think this is fine for now as they are fairly vague.

dallasread:
There is a bit of a conundrum here as to what we display by default. This PR prioritizes:

- Is there a search term that matches the URL nicely?
  - If yes, show it.
  - If not, are there any recent articles?
    - If yes, show them.
    - If not, is there a getting started?
      - If yes, show it.

Have a touch more work to do on this before it's ready.

PR #1429, opened 2025-03-17
Fix support widget UI defects

kojdm:
Fixes the following UI defects:

- [x] Alert icons aren't displayed correctly when an article is rendered directly on the modal search.
- [x] Issue rendering dynamic content in search results

Fixes https://github.com/dnsimple/dnsimple-app/issues/29898

## QA

- Searching for 'import' renders the article inside the modal and the alert icons are clearly visible.
  image

- Searching for 'located' does not render the variable in the excerpt
  - Our deployment previews currently pull in the `search.json` data from production, so an alternative way to test this would be to visit https://deploy-preview-1429--dnsimple-support.netlify.app/articles/where-are-your-servers-located/ and verify that the metadata excerpt is rendered correctly.
  image

### DIFF
--- a/content/articles/where-are-your-servers-located.md
+++ b/content/articles/where-are-your-servers-located.md
@@ -1,6 +1,6 @@
 ---
 title: Where are your servers located?
-excerpt: Our Anycast network consists of <%= POPS.count %> points-of-presence around the world.
+excerpt: Our Anycast network consists of multiple points-of-presence around the world.
 categories:
 - DNSimple
 ---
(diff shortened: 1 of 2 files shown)


content/articles/where-are-your-servers-located.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Where are your servers located?
-excerpt: Our Anycast network consists of <%= POPS.count %> points-of-presence around the world.
```
kojdm, line 3: I couldn't find a way to get erb to work inside the frontmatter. I believe it should be reserved only for static data. FWIW, this doesn't seem to work in our compiled site's metadata either:

image

dallasread: 😄

dallasread (approved):

PR #1449, opened 2025-04-14
Ignore modal key events when modal is not opened

juankuquintana:
This PR adds a verification to return early on the widget modal when the modal is not opened.

## :mag: QA

- Open widget via command
- Search on widget

## :shipit: Verification

- [ ] QA

### DIFF
--- a/_widget/src/main.js
+++ b/_widget/src/main.js
@@ -1,7 +1,5 @@
 import initialize from './initialize.js';
 import storeRecentlyVisited from './recently-visited.js';
 
-const elementId = 'dnsimple-support-widget';
-
 const app = initialize(document);
 storeRecentlyVisited(app);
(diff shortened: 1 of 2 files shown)


_widget/src/main.js
```diff
@@ -1,7 +1,5 @@
 import initialize from './initialize.js';
 import storeRecentlyVisited from './recently-visited.js';
 
-const elementId = 'dnsimple-support-widget';
```
sbastn, line 4: Removed because it's not used?

sbastn (approved):
✅ I tried to break it, but I couldn't

_widget/src/main.js
juankuquintana: Yep

PR #1454, opened 2025-04-16
Update Creating Accounts for Clients

itsalyse:
Updates https://support.dnsimple.com/articles/setting-up-accounts-for-clients/ to include the correct, current article for managing seats.

### DIFF
--- a/categories/account.yaml
+++ b/categories/account.yaml
@@ -25,7 +25,7 @@ Account members:
 - "Domain Access Control"
 - "Managing seats"
 - "Managing Multiple Members on One Account"
-- "Creating accounts for clients"
+- "Creating Accounts for Clients"
 - "Differences between Account Collaborator and Multi-User Accounts"
 
 Identity Providers:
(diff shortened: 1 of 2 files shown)


stephanoodle:
> This was the original article linked https://support.dnsimple.com/articles/account-users/#adding-members-to-an-account, which is the outdated version of https://support.dnsimple.com/articles/managing-seats/
> 
> Are these meaningfully different enough to justify keeping both?

This could be one article, IMO. Managing Seats is the one I would add some of the adding members to an account information (such as the extra steps to enter an email to invite a user, and also the access levels). I think having two articles is silly, and neither gives all of the relevant information, but if they were merged, the doc would have all of the information.

stephanoodle (approved):

itsalyse:
Okay, I'll update the link for this article so it at least goes to the place with the correct steps for adding a seat, and then draft a version of the managing seats article that pulls the still-relevant info from the adding members article

Pemacaa (approved):

PR #1469, opened 2025-04-24
Updating title of SOA record

dperrymorrow:
for keyword ranking, adding DNS prior to SOA in the H1 and h2


laetitiaeden (approved):
This is an experiment to see if we can rank back around the notion of SOA by adding DNS in front of it.

PR #1471, opened 2025-04-25
Update dictionary.js

Pemacaa:
Update the dictionary so that Managing Seats brings up the Managing Seats support page.

Searching for managing does not capture Managing Seats support.


dallasread (approved):
🙌 Thanks!

PR #1476, opened 2025-04-30
Fixing internal redirects

dperrymorrow:
the support site redirects to a trailing slash url if the url does not have a trailing slash. This makes many of our internal links look like 301 redirects to Google Crawler. 

This pr addresses these urls adding a trailing slash so that they do not redirect.


laetitiaeden (approved):
Fix issue that Siteguru pointed out

sbastn (approved):
Awesome!

PR #1496, opened 2025-05-21
Fixing lack of trailing slashes on urls

dperrymorrow:
urls without trailing slashes are seen as an internal redirect. This PR addresses that.


dallasread (approved):
LGTM.

I wonder if we could create a jekyll/nanoc rule to enforce this.

PR #1497, opened 2025-05-22
Adjust the scroll margin to show the section header

sbastn:
When navigating via the table of contents or an anchor link, section headings were disappearing under our sticky header. This PR adds a top scroll margin to each heading so that, when you jump to a section, its title is fully visible below the header.

There is good support for [scroll-margin](https://caniuse.com/?search=scroll-margin).

### Before

https://support.dnsimple.com/articles/secondary-dns/#getting-started

Screenshot 2025-05-22 at 11 43 40

### After

https://deploy-preview-1497--dnsimple-support.netlify.app/articles/secondary-dns/#getting-started

Screenshot 2025-05-22 at 11 43 18

### DIFF
--- a/content/assets/css/_typography.scss
+++ b/content/assets/css/_typography.scss
@@ -17,6 +17,7 @@ body {
 
 h1, h2, h3 {
   font-weight: normal;
+  scroll-margin-top: 65px;
 }
 
 h1 {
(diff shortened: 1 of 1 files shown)


dallasread (approved):
😎

PR #1498, opened 2025-05-22
Shared Domains deprecation

dallasread:
Since we've had this feature deprecated for some time, we're removing mentions of it in the app. It seems to make sense to fall a suit on the support site as well.

Belongs to https://github.com/dnsimple/dnsimple-app/issues/27411

### DIFF
--- a/_widget/src/components/app/dictionary.js
+++ b/_widget/src/components/app/dictionary.js
@@ -13,7 +13,6 @@ export default {
   "cancel": "unsubscribe",
   "delete account": "unsubscribe",
   "letsencrypt": "lets encrypt",
-  "membership": "share management",
   "team": "member",
   "remove": "delete",
   "registration": "register",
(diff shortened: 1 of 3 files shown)


categories/domains.yaml
```diff
@@ -19,7 +19,6 @@ Expiration:
 Managing domains:
 - "Domain Access Control"
 - "Organizing and Labeling Domains"
-- "Sharing Domain Management with Other Users"
```
aeden, line 22: Why remove this rather than having it link to the Domain Access Control article?

aeden (commented):
Left a question.

categories/domains.yaml
dallasread: It seemed weird to leave a page about a deprecated feature around. I took the route of removing the page, but putting in a redirect to keep the search engines happy. Ultimately, I'm ok with either way - just wanted to get the convo started.

dallasread: @Pemacaa @stephanoodle Thoughts?

Pemacaa: From time to time, we get customers who are confused because they can't access domains they previously managed. When that happens, I always point them to the shared domain management article so it's a nice to have. That said, it would be even more useful if the article **briefly** explained when the old shared domain feature was deprecated and how it used to work since the current article explains the present process.

stephanoodle (approved):

PR #1502, opened 2025-05-28
Search: Improve `dns management` and `manage account`

dallasread:
This PR removes `dns management` from being a target search term and adds `manage account`.

You can review the new results @ https://deploy-preview-1502--dnsimple-support.netlify.app/.

### DIFF
--- a/spec/_widget/components/search.spec.js
+++ b/spec/_widget/components/search.spec.js
@@ -82,7 +82,7 @@ describe('Search', () => {
         'Managing A Records': 5,
         'What\'s an A Record?': 1
       },
-      'account management': {
+      'manage account': {
         'Account Creation': 1,
         'Account Activation': 5
       },


spec/_widget/components/search.spec.js
```diff
@@ -198,7 +202,7 @@ describe('Search', () => {
         'Forgot Password': 1
       },
       'email not working': {
-        'What\'s an MX Record?': 4,
+        'What\'s an MX Record?': 5,
```
dallasread, line 205: If you're wondering how I come up with these out-of-nowhere numbers, when I run the tests (`yarn test`), sometimes they fail:

image

And we also have a warning that indicates that we can actually expect the article in the results higher than it is:

image

stephanoodle (approved):

itsalyse (approved):
Thanks, Dallas!

PR #1510, opened 2025-06-10
Add note for catch-all email forwards

lokst:
Add a note for catch-all email forwards


dallasread (approved):

dallasread:
The idea is to document the existing behaviour because it is not described anywhere at the moment. When the new changes drop in a month, we'll update it.

stephanoodle (approved):
I'm not sure how many customers will notice this, but I think it's good information that we probably should have had in the documentation long ago. It will also help to have it to explain a bit when the email about the change goes out.

PR #1513, opened 2025-06-23
Create enabling-dnssec.md

stephanoodle:
Adding a new article,  Enabling DNSSEC, to the Support site.

### DIFF
--- a/content/articles/enabling-dnssec.md
+++ b/content/articles/enabling-dnssec.md
@@ -1,0 +1,43 @@
+---
+title: Enabling DNSSEC
+excerpt: Enable DNSSEC for your domain. Understand the prerequisites, follow the configuration steps, and know what to expect after activation.
+categories:
+- DNSSEC
+---
+
+# Enabling DNSSEC
+
+## Prerequisites
+
+You cannot enable DNSSEC if you have set up Secondary DNS. They will not work in conjunction. Ensure you are not currently using Secondary DNS, or disable Secondary DNS before using DNSSEC. You can read more about why here.
+
+1.  Use the **account switcher** at the top of the page to select the appropriate account.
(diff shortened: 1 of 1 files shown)


stephanoodle:
account switcher

stephanoodle:
@dperrymorrow I think the checks are failing due to SEO ranking? 

I also need some help with screenshots. If you could walk me through the process, I think I'll be able to handle it.

itsalyse:
@stephanoodle success 🥳

dperrymorrow:
> It looks like it's looking for the article to be in a certain position in our search widget: https://github.com/dnsimple/dnsimple-support/blob/main/spec/_widget/components/search.spec.js#L89-L92


Yes the specs for the search widget, are asserting on the number of results for a topic it looks like. When you add a new article in a category, the specs will fail. I can talk to @dallasread about how we could make this less brittle, but I can fix this in the meantime.

spec/_widget/components/search.spec.js
```diff
@@ -87,8 +87,8 @@ describe('Search', () => {
         'Account Activation': 5
       },
       'enable dnssec': {
-        'DNSSEC': 1,
-        'Why DNSSEC and Secondary DNS may not work together': 3
+        'DNSSEC': 2,
```
dallasread, line 90: @dperrymorrow This is the minimum rank for this article/search term (not # of articles). I've kept them relatively tight so that we can notice larger changes. These are not an exhaustive list of results for the term, but the articles that we've said MUST show up for the search term.

If you run the specs, you'll see warnings for search terms where we can tighten up the results (if there are any).

I recommend we add `Enabling DNSSEC` here as a `1` as well.

dperrymorrow (approved):

PR #1517, opened 2025-06-24
Rotate DNSSEC Keys article

Pemacaa:
Belongs to https://github.com/dnsimple/dnsimple-customer-success/issues/7

### DIFF
--- a/spec/_widget/components/search.spec.js
+++ b/spec/_widget/components/search.spec.js
@@ -89,9 +89,8 @@ describe("Search", () => {
       'enable dnssec': {
         'Disabling DNSSEC': 1,
         'DNSSEC': 3,
-        'DNSSEC Glossary': 5,
-        'Using Cloudflare DNSSEC with DNSimple': 6,
-        'Why DNSSEC and Secondary DNS may not work together': 7,
+        'DNSSEC Glossary': 6,
+        'Using Cloudflare DNSSEC with DNSimple': 7,
       },
       "creating alias record": {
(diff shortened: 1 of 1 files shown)


content/articles/rotate-dnssec-key.md
```diff
@@ -0,0 +1,66 @@
+--
```
stephanoodle, line 1: Should be --- to format the YAML correctly.

content/articles/rotate-dnssec-key.md
```diff
@@ -0,0 +1,66 @@
...
+title: Rotate DNSSEC Keys
+excerpt:TBD
```
stephanoodle, line 3: excerpt: excerpt: Guides DNSSEC key rotation for DNSimple users, detailing auto-rotation and essential manual steps to maintain domain security.

what do you think?

stephanoodle (changes requested):
Very nice. I see you were able to figure out the screenshots! 

Once the 2 things I commented on are addressed, I would say this is good to go. I believe we need Alyse to review one more time now that it's in a PR, and then someone from marketing (probably Dave?).

Pemacaa:
> I believe we need Alyse to review one more time now that it's in a PR, and then someone from marketing (probably Dave?)

Yes, absolutely! As I mentioned earlier, I just needed to do a bit of cleanup first so Alyse wouldn’t run into too much of a mess 😄

dperrymorrow:
fixed the test suite again after merging main down, preview should build now

PR #1537, opened 2025-07-07
Create what-are-ds-records.md

itsalyse:
Adds an article detailing what DS records are and why they're important 

Belongs to https://github.com/dnsimple/dnsimple-customer-success/issues/12

### DIFF
--- a/content/articles/what-are-ds-records.md
+++ b/content/articles/what-are-ds-records.md
@@ -57,8 +57,9 @@ Here's how it works:
     - Fetches the DNSKEY from your domain zone
     - Compares the DS digest to your DNSKEY
 
-The resolver creates a digest of the DNSKEY using the same hash algorithm and digest type specified in the DS record. Even though the DS record doesn't contain the full public key, it includes a cryptographic hash of it. By hashing the DNSKEY locally and comparing it to the digest in the DS record, the resolver can confirm the two are linked.
-If the created digest matches the DS record, the resolver trusts your DNSKEY and the DNS records it signs. If they don't match — for example, if you've rotated your KSK but haven't updated the DS record — resolvers won't trust your domain's DNSSEC signatures, and your domain may become unreachable.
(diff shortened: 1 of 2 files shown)


stephanoodle (approved):

Pemacaa (approved):

content/articles/what-are-ds-records.md
```diff
@@ -0,0 +1,138 @@
...
+
+1. **Resolver validates the Chain of Trust**
+    When a DNSSEC-aware resolver queries your domain, it:
+    - Retrieves the DS record from the parent zone
+    - Fetches the DNSKEY from your domain zone
+    - Compares the DS digest to your DNSKEY
+
+The resolver creates a digest of the DNSKEY using the same hash algorithm and digest type specified in the DS record. Even though the DS record doesn't contain the full public key, it includes a cryptographic hash of it. By hashing the DNSKEY locally and comparing it to the digest in the DS record, the resolver can confirm the two are linked.
+
+If the created digest matches the DS record, the resolver trusts your DNSKEY and the DNS records it signs. If they don't match — for example, if you've rotated your KSK but haven't updated the DS record — resolvers won't trust your domain's DNSSEC signatures, and your domain may become unreachable.
```
ggalmazor, line 62: > for example, if you've rotated your KSK but haven't updated the DS record

This is not a good example because the key rotation would create a new set of DS/DNSKEY records with a different key tag (resolvers will try to verify DS&DNSKEY records that match the same key tag).

A possible scenario for this would be a copy&paste error while provisioning the DS record in the parent zone, or malicious third-party tampering with DNS answers for the child zone.

ggalmazor: Sorry I didn't catch it earlier! 😅 I'll drop a suggestion in the doc

Pemacaa:
@itsalyse, the three remaining screenshots have been added:

Named:
Add DS record (Key-Data interface)
Delete DS records
Delete DS records 2

PR #1541, opened 2025-07-09
Create dnssec-at-dnsimple.md

itsalyse:
Creates a pillar page linking out to the various articles on DNSSEC at DNSimple. The goal of this page is to provide a TL;DR of each article and make it easy for users (especially those new to DNSSEC) to find what they're looking for.

### DIFF
--- a/content/articles/dnssec-at-dnsimple.md
+++ b/content/articles/dnssec-at-dnsimple.md
@@ -1,0 +1,53 @@
+---
+title: DNSSEC at DNSimple: Everything You Need to Know
+excerpt: Learn where to find everything you need to know about DNSSEC at DNSimple.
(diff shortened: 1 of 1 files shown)


stephanoodle:
I forgot to put "[Using Cloudflare DNSSEC with DNSimple](https://support.dnsimple.com/articles/cloudflare-ds-record/)" in the pillar page altogether. 

Adding it now.

Pemacaa (approved):

stephanoodle:
@dperrymorrow two things - this is supposed to replace https://support.dnsimple.com/articles/dnssec/ (the content on that page is in other articles in the section). We also need to remove (or update to this) that content. 

And then obviously this is choking on search.

dperrymorrow (approved):

PR #1547, opened 2025-07-15
Making the search widget specs more forgiving

dperrymorrow:
### Problem:
- Each time an article is added to a category the search widget specs fail as it will change the order of the articles returned.
- Each time an article's title is changed the search specs will fail as it is looking for the actual `String` of the articles title to match. 

> This problem makes it nearly impossible for @stephanoodle, @Pemacaa  or @itsalyse to author new articles or change titles using the GitHub in browser editor, without the support of a developer to come change the expected order, Strings of titles to match what the search spec is expecting. 

### Proposed Solution:
- Make assertions that the desired articles are merely in the list of results, not the order of them. 
- Make assertions based on the `id` of the article _(the filename of the `.md` file)_ rather than the title of the article in the Frontmatter.  This lets you assert on the article without being bound to the `String` of the title being an exact match.


dallasread (approved):
I'd suggest that it's a good idea for us to know how & when we are negatively impacting our search results. The code looks fine though – no point me being the blocker on this.

dperrymorrow:
> I'd suggest that it's a good idea for us to know how & when we are negatively impacting our search results. The code looks fine though – no point me being the blocker on this.

I would agree, and understand the intention of the specs, but its literally every time an article is edited / created in one of the spec categories, the specs will fail.

I am open to other ideas that are less drastic than my proposed solution above.

PR #1605, opened 2025-10-07
Announcement non-compliant ENT deprecation

weppos:
This article inform affected customers about the deprecated ENT behavior.

Belongs to https://github.com/dnsimple/dnsimple-business/issues/2229
Belongs to https://github.com/dnsimple/dnsimple-business/issues/2290

### DIFF
--- a/content/articles/announcement-change-ent-behavior.md
+++ b/content/articles/announcement-change-ent-behavior.md
@@ -1,0 +1,75 @@
+---
+title: Change of non-compliant ENT behavior
+excerpt:
+categories:
+---
+
+# Change of non-compliant ENT behavior
+
+<info>
+This article contains information that is only relevant to customers with specific zone records. If you are impacted, you will receive an email from us asking you to take action.
+</info>
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+As part of our ongoing commitment to DNS standards compliance and improved resolution behavior, we are updating our name server behavior to properly handle Empty Non-Terminal (ENT) records according to RFC 4592. This change will be implemented on **December 1, 2024**, with a transition period ending on **November 30, 2024**.
+
+Following this update, DNSimple name servers will respond correctly to ENT queries, returning `NOERROR` with no data instead of incorrectly applying wildcard records. This change ensures full compliance with DNS standards and improves predictability for DNSSEC-compliant zones.
+
+## Who's impacted?
+
+This announcement impacts all customers with zones that contain an Empty Non-Terminal (ENT) record.
+
+## What are Empty Non-Terminals?
+
+In DNS, an *Empty Non-Terminal* is an intermediate domain name in a DNS hierarchy that exists only because it is a path to a deeper name, but has no records of its own. For example, in the following zone `b.c.example.com` is an Empty Non-Terminal. It has no records itself, but exists because of `a.b.c.example.com`.
+
+```
+$ORIGIN example.com.
+*.c.example.com.     IN  CNAME   one.test.
+a.b.c.example.com.   IN  CNAME   two.test.
+```
+
+A typical example is represented by [DKIM records](https://support.dnsimple.com/articles/dkim-record/), that often generates an ENT due to its specific naming structure.
+
+## What's changing
+
+Let's take the previous zone as reference:
+
+- `b.c.example.com` is an Empty Non-Terminal. It has no records itself, but exists because of `a.b.c.example.com`.Add a comment on lines R34 to R35Add diff commentMarkdown input:  edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a reviewReturn to code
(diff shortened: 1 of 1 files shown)


weppos:
Tagging for review:

- @itsalyse for copy support
- @DXTimer for technical cross-review

content/articles/announcement-change-ent-behavior.md
```diff
@@ -0,0 +1,75 @@
...
+</info>
+
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+As part of our ongoing commitment to DNS standards compliance and improved resolution behavior, we are updating our name server behavior to properly handle Empty Non-Terminal (ENT) records according to RFC 4592. This change will be implemented on **December 1, 2024**, with a transition period ending on **November 30, 2024**.
```
itsalyse, line 20: @weppos I think these dates are wrong

stephanoodle: rfc link should be added here as well - https://datatracker.ietf.org/doc/html/rfc4592#section-4.9

weppos: 🤦 The dates are correct the year is wrong. It turns out we're already in 2025, not 2024.

content/articles/announcement-change-ent-behavior.md
```diff
@@ -0,0 +1,76 @@
...
+
+## What's changing?
+
+Let's take the previous zone as reference:
+
+- `b.c.example.com` is an Empty Non-Terminal. It has no records itself, but exists because of `a.b.c.example.com`.Add a comment on lines R34 to R35Add diff commentMarkdown input:  edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a reviewReturn to code
+- A query to `b.c.example.com` is expected to return NOERROR with no data (`NODATA`), indicating that the name exists but has no records. Importantly, the wildcard at `*.c.example.com` does not apply here, because the name exists (even if empty).
+- On the other hand, a query to something like `x.c.example.com` would be synthesized from the wildcard, returning a CNAME to `one.test`.
+
+As of today, DNSimple name server response is not compliant with ENT RFC 4592. When querying for the ENT, DNSimple name servers would incorrectly return the wildcard.
```
DXTimer, line 48: The first bullet point is corrupted, and the NODATA description is slightly off. 

```suggestion
- `b.c.example.com` is an Empty Non-Terminal. It has no records itself, but exists because of `a.b.c.example.com`.
- According to RFC 4592, a query for `b.c.example.com` should return `NOERROR` with an empty answer section (also known as a `NODATA` response), indicating that the name exists but has no records. Importantly, the wildcard at `*.c.example.com` should not apply here, because the name exists (even if empty).
- On the other hand, a query for a name like `x.c.example.com` (which doesn't exist) should be synthesized from the wildcard, returning a CNAME to `one.test`.

Currently, DNSimple name servers do not comply with RFC 4592 for ENTs. When querying an ENT that is a descendant of a wildcard, DNSimple name servers incorrectly return the wildcard match.
```

content/articles/announcement-change-ent-behavior.md
```diff
@@ -0,0 +1,76 @@
...
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+As part of our ongoing commitment to DNS standards compliance and improved resolution behavior, we are updating our name server behavior to properly handle Empty Non-Terminal (ENT) records according to [RFC 4592 section 4.9](https://datatracker.ietf.org/doc/html/rfc4592#section-4.9). This change will be implemented on **December 1, 2025**, with a transition period ending on **November 30, 2025**.
+
+Following this update, DNSimple name servers will respond correctly to ENT queries, returning `NOERROR` with no data instead of incorrectly applying wildcard records. This change ensures full compliance with DNS standards and improves predictability for DNSSEC-compliant zones.
```
DXTimer, line 22: Fixing NODATA definition.


```suggestion
Following this update, DNSimple name servers will respond correctly to ENT queries, returning `NOERROR` with an empty answer section (a `NODATA` response) instead of incorrectly applying wildcard records. This change ensures full compliance with DNS standards and improves predictability for DNSSEC-compliant zones.
```

content/articles/announcement-change-ent-behavior.md
```diff
@@ -0,0 +1,76 @@
...
+Currently, DNSimple name servers do not comply with RFC 4592 for ENTs. When querying an ENT that is a descendant of a wildcard, DNSimple name servers incorrectly return the wildcard match.
+
+For instance, querying `b.c.example.com` would return:
+
+```
+$ dig @127.0.0.1 b.c.example.com A +short
+one.test.
+```
+
+Once the correct behavior is rolled out, DNSimple name servers will respons with `NOERROR` and an empty response (no records).
```
DXTimer, line 57:
```suggestion
Once the correct behavior is rolled out, DNSimple name servers will respond with `NOERROR` and an empty answer section (no records).
```

weppos:
> From a technical stand point the contents look good.

Thanks for the review. I applied the changes. Let's have a final review from @itsalyse and @stephanoodle.

PR #1607, opened 2025-10-08
Bump rack to latest version (CVE-2025-61770)

jacegu:
Addresses dependabot security alerts:
- https://github.com/dnsimple/dnsimple-support/security/dependabot/104
- https://github.com/dnsimple/dnsimple-support/security/dependabot/103
- https://github.com/dnsimple/dnsimple-support/security/dependabot/102

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -115,7 +115,7 @@ GEM
     puma (7.0.4)
       nio4r (~> 2.0)
     racc (1.8.1)
-    rack (3.1.16)
+    rack (3.2.2)
     rack-livereload (0.5.1)
       rack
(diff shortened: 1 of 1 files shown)


kntsoriano:
@jacegu This is an example of an automated update created by dependabot: https://github.com/dnsimple/dnsimple-developer/pull/869

In this case, dependabot bumped the version to 3.1.17 instead of 3.2.2.

There is also an open PR from dependabot doing the same change: https://github.com/dnsimple/dnsimple-support/pull/1606. This also address the 3 alerts as linked. I think that we should merge  https://github.com/dnsimple/dnsimple-support/pull/1606 instead of this one.

PR #1648, opened 2025-11-14
Update to explain how to find te domain push identifier

Pemacaa:
I added a short clarification in the “Initiating a transfer” section to help customers understand where to find the domain push identifier.


dallasread (approved):
Thank you!

PR #1659, opened 2025-11-19
Update dns.yaml

stephanoodle:
Making an announcements section in DNS to fix orphan pages.


dperrymorrow:
> I'm using the exact article names, but for some reason, they aren't showing on the page.

I have pushed the fix. The articles you wanted to appear on the page needed to belong to the DNS category in order to be used on the category page.

stephanoodle:
Oh! That makes complete sense and I can't believe I forgot about that. Thank you @dperrymorrow

PR #1664, opened 2025-11-21
Enterprise Category

stephanoodle:
Creating an Enterprise Category and then adding relevant articles to it for the Revenue Team

### DIFF
--- a/categories/enterprise.yaml
+++ b/categories/enterprise.yaml
@@ -43,7 +43,7 @@ General:
   - "API Rate Limit"
   - "API Sandbox"
   - "Terraform Provider"
-  - "Domain Zone Files"
+  - "What Is a Zone File?"
   - "What is a TLD?"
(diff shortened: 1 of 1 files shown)


categories/enterprise.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - "What Are CDS and CDNSKEY?"
+  - "SSL/TLS Certificates"
+  - "SSL Certificate Types"
+  - "Ordering a Let's Encrypt Certificate"
+  - "Ordering a Wildcard SSL Certificate"
+
+Infra:
+  - "DNSimple Plans"
+  - "Where are your servers located?"
+  - "Managing Vanity Name Servers"
```
rufizdnsimple: I think it makes sense to reverse the order of these articles

felipednsimple (approved):

PR #1671, opened 2025-11-25
Rename `domain push identifier` to `account identifier`

dallasread:
This PR renames `domain push identifier` to `account identifier`.


Pemacaa:
Thank you @dallasread It looked good!

PR #1675, opened 2025-11-28
Name Servers section rewrite

aeden:
Start with the content from all articles in the Name Servers and replace the existing articles with new articles using the Diataxis framework (https://diataxis.fr/) and following the voice from articles in the DNSSEC and DNS category, . Produce a list of gaps of missing articles for the Name Servers category that should be added in the future based on functionality provided by DNSimple related to name servers and delegating domains. Create a new pillar page for the Name Servers category using https://support.dnsimple.com/articles/dnssec/  as an example template.


stephanoodle:
closing this ticket as it has been split out into smaller PRs to deal with

PR #1678, opened 2025-12-02
Add article about Empty Non Terminals

weppos:
Given the complexity of the topic, and our changes to be RFC-compliant, I feel like we want to have a permanent article about ENTs and solutions.

### DIFF
--- a/categories/dns.yaml
+++ b/categories/dns.yaml
@@ -13,6 +13,7 @@ Explanation:
   - What Is a DKIM Record?
   - What Is a DMARC Record?
   - What Are DS Records?
+  - What Are Empty Non-Terminals (ENT)?
   - What Are Glue Records?
   - What Are HINFO Records?
   - What Is an MX Record?
@@ -70,6 +71,7 @@ How to:
   - Export Zone File
   - How To Add Common DNS Records
   - How to Change TTL in DNSimple
+  - How to Fix Empty Responses from Empty Non-Terminals
   - How to Use the Record Editor
   - How to Use dig
   - How to Use CoreDNS as an Integrated DNS Provider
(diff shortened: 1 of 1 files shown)


stephanoodle:
@weppos, we need to have this Diataxis compliant.

This means breaking up the main article into:

- Explanation: What Are Empty Non-Terminals (ENTs)?
- How to: How to Fix Empty Responses from Empty Non-Terminals
- How to: Troubleshooting Empty Non-Terminal Issues

content/articles/empty-non-terminals.md
```diff
@@ -0,0 +1,377 @@
...
+### Table of Contents {#toc}
+
+* TOC
+{:toc}
+
+---
+
+## What are Empty Non-Terminals?
+
+In DNS, an **Empty Non-Terminal** (ENT) is an intermediate domain name in a DNS hierarchy that exists only because it is a path to a deeper name, but has no records of its own.
```
Pemacaa: One small thought from a reader's perspective. I’m still finding this explanation it a bit abstract on a first read, especially for users who aren’t already familiar with the concept. Speaking also for myself, it took me a while to understand why ENT exists.

What do we think of adding more info?


```suggestion
Empty Non-Terminals exist to explain why DNS sometimes behaves in a way that feels unexpected. 

Customers don't create Empty Non-Terminals on purpose. Instead, they appear as a side effect when customers add normal DNS records for things like websites, email authentication, or SSL certificates. You usually only notice ENTs when a DNS query returns no records, or when a wildcard record stops matching as expected.

In DNS, an **Empty Non-Terminal** (ENT) is an intermediate domain name in a DNS hierarchy that exists only because it is a path to a deeper name, but has no records of its own.
```

Pemacaa (approved):

stephanoodle (approved):

PR #1682, opened 2025-12-03
Update what-is-zone-file to fix 404 issue

stephanoodle:
The URL was changed for the zone file documentation - need to update the redirect to fix 404 issue. 

Closes https://github.com/dnsimple/dnsimple-business/issues/2372


stephanoodle:
> Is that sufficient to redirect correctly? I have a change as well locally and that change updated _redirects in the repo root.

I missed a - that I fixed, but that is how multiple redirects are done on the SSL Email Validation article.

If you have a better way, by all means

PR #1752, opened 2026-01-20
Fixing Callouts in Some Documentation

stephanoodle:
Some documentation was written in the transition and callout formatting was missed.

Some callouts also were not working due to spacing/punctuation when manually updating.

### DIFF
--- a/content/articles/disabling-dnssec.md
+++ b/content/articles/disabling-dnssec.md
@@ -26,8 +26,8 @@ categories:
 
 ## Important timing considerations
 
-> [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


content/articles/rotate-dnssec-key.md
```diff
@@ -31,7 +31,7 @@ You will need to manually manage your DS records at your domain registrar.
 1. Once our system verifies the old DS record is no longer present in the parent zone, we remove the old key set, and the key set rotation is completed.
 
 > [!IMPORTANT]
```
weppos: It may not be the case, but FYI you can also write it as:


```suggestion
> [!IMPORTANT] TTL considerations
> After you add the DS records at your registrar, it can take up to 24 hours for the key rotation process to complete. This is because the TTL (Time-to-Live) for DS records is 24 hours, and our system needs to verify that the new DS records have fully propagated before proceeding with the rotation. The rotation will not complete until the TTL has expired and the new DS records are fully present in the parent zone.
```

PR #1753, opened 2026-01-20
Docs: How to tweak widget results

dallasread:


stephanoodle:
@dallasread can you provide more detail on how to do these? 

I am already familiar with the ## Search Dictionary thanks to the wiki docs. 

But the rest? I'm a bit lost on. For instance, I'm fairly sure I want to use ## Priority Files, but I don't understand how to get to where to do that - priorities don't seem to exist where this says. 

I beat my head against this, and attempted with Cursor as well, and I couldn't get anywhere (so I wasn't able to fix my PR, and now another PR is erroring).

PR #1958, opened 2026-05-27
Fix sitemap.xml 500 error

stephanoodle:
## What and why

`https://support.dnsimple.com/sitemap.xml` was returning a **500 Internal Server Error**, making the sitemap unreachable by search engines (Googlebot, Bingbot, etc.) despite being correctly declared in `robots.txt`.

The root cause was the bare `xml_sitemap` call in `content/sitemap.xml.erb` with no arguments:

```erb
<%= xml_sitemap %>
```

When called without filtering, nanoc's `XMLSitemap` helper attempts to include **every item in the site** — articles, binary image files, CSS, JS assets, and the sitemap file itself. Including binary items and the sitemap's own item causes a recursive dependency error during compilation: the sitemap tries to check whether its own compiled snapshot is binary in order to compile itself, which creates an unresolvable circular dependency. This results in the sitemap either not being generated correctly or being served as a broken file, producing the 500.

## The fix

Filter the items passed to `xml_sitemap` to only include pages that route to `.html`:

```erb
<%= xml_sitemap(items: @items.select { |i| i.path&.end_with?('.html') }) %>
```

The safe-navigation operator (`&.`) handles items with no path (binary files, unrouted assets) gracefully so they are excluded before the `end_with?` check runs. This matches the pattern recommended in the [nanoc XMLSitemap helper docs](https://nanoc.app/doc/reference/helpers/#xmlsitemap) and resolves the recursive dependency that was causing the 500.

## What's included

All 501 support articles will appear in the sitemap. Binary assets, CSS, JS, and the sitemap file itself are excluded.

## References

- nanoc issue [#1082](https://github.com/nanoc/nanoc/issues/1082) — documents the exact recursive dependency failure mode
- [nanoc XMLSitemap helper docs](https://nanoc.app/doc/reference/helpers/#xmlsitemap)

### DIFF
--- a/content/sitemap.xml.erb
+++ b/content/sitemap.xml.erb
@@ -1,1 +1,1 @@
-<%= xml_sitemap %>
+<%= xml_sitemap(items: @items.select { |i| i.path&.end_with?('.html') }) %>
--- a/content/sitemap.xml.erb
+++ b/content/sitemap.xml.erb
@@ -1,1 +1,1 @@
-<%= xml_sitemap(items: @items.select { |i| i.path&.end_with?('.html') }) %>
+<%= xml_sitemap(rep_select: ->(rep) { rep.path == "/" || rep.path.end_with?("/") }) %>


weppos:
Please let's verify that whatever is the correct solution, works and is applied consistently both here and in developer site. They are sibling projects and must be kept in sync.

sbastn:
I cannot reproduce the 500 error, but I do see all the assets: pages + screenshots. The date seems to be off as well because they all have the same one. Probably from the last deploy.

Screenshot 2026-05-27 at 11 12 16

sbastn:
The sitemap on this current PR is empty:

Screenshot 2026-05-27 at 11 15 49

sbastn:
I've pushed a fix to filter by path instead of HTML extension:
Screenshot 2026-05-27 at 11 39 34

sbastn:
We should consider removing the `lastmod` attribute. Google has explicitly said they ignore or distrust it when it is inaccurate or automatically updated without meaningful page changes. A bad `lastmod` can be worse than omitting it entirely.

However, we can start by making this change, and remove the `lastmod` in a different PR.

sbastn (approved):

PR #1964, opened 2026-05-27
Fix duplicate Troubleshooting key and sync Name Servers category YAML

stephanoodle:
## Summary

- Fixes a duplicate `Troubleshooting:` top-level key in `categories/name_servers.yaml` — invalid YAML where the second entry silently overwrote the first
- Merges both Troubleshooting subsections (`After delegation`, `DNSimple name servers`) under a single key
- Syncs entries added by recently merged PRs: adds `Name Server Management at DNSimple` to Getting started, and `Interface reference`, `Glossary`, and `NS record` subsections to Reference (#1939, #1937, #1938)

## Test plan

- [x] Verify https://support.dnsimple.com/categories/name-servers/ renders all sections correctly with no missing or duplicated articles

### DIFF
--- a/categories/name_servers.yaml
+++ b/categories/name_servers.yaml
@@ -26,10 +26,10 @@ How-to:
     - "Manage Vanity Name Servers"
 
 Troubleshooting:
-  DNSimple name servers:
-    - "Troubleshoot DNSimple Name Servers"
   After delegation:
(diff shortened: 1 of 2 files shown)


stephanoodle:
Replacing with a clean branch off main.

PR #1983, opened 2026-06-10
Split regional record regions from PoP list (fixes regional-records article)

stephanoodle:
## Summary

PR #1754 expanded the shared `POPS` constant to the full 16-location Anycast edge network. `POPS` is also rendered by the **Regional Records** article to tell customers which regions they can pin a record to — but the regional records feature only accepts the **8** regions in `Dnsimple::Region::ALL` (dnsimple-app). The expansion made the article advertise regions the app/API silently rejects, and dropped Ashburn (IAD), which is valid. Reported in **ticket 546048**.

This PR separates the two concepts and brings the affected articles up to the repo's writing/SEO rules.

- Add a dedicated `REGIONAL_RECORD_REGIONS` constant (the 8 selectable regions) and point the Regional Records article at it; `POPS` still drives the two Anycast articles.
- Rename the article to **"Regional Records Explained"** (keyword-leading title); update category YAMLs and the `dns.md` cross-link; remove pre-existing duplicate "Regional Records" listings.
- SEO/GEO + structure fixes on the article: `meta`/`excerpt`, table of contents, PoP cross-link, heading/spacing.
- Define "point of presence": new glossary entry in `name-servers-glossary`, and promote `anycast.md` from a bare list into an Explanation that defines a PoP and contrasts it with regional records.

## Files changed

- `lib/default.rb` — new `REGIONAL_RECORD_REGIONS` constant
- `content/articles/regional-records.md` — uses new constant, retitled, SEO/GEO + structure
- `content/articles/anycast.md` — promoted to Explanation defining a PoP
- `content/articles/name-servers-glossary.md` — new "Point of Presence (PoP)" entry
- `content/articles/dns.md` — cross-link text + duplicate removed
- `categories/dns.yaml`, `categories/enterprise.yaml` — rename + duplicate removed

## ⚠️ Follow-up: cross-repo coupling (not addressed here)

`REGIONAL_RECORD_REGIONS` is a hand-maintained copy of `Dnsimple::Region::ALL` ([dnsimple-app `lib/dnsimple/region.rb`](https://github.com/dnsimple/dnsimple-app/blob/main/lib/dnsimple/region.rb)), which is the real source of truth (`Region.filter` rejects anything outside it). The two repos are coupled only by meaning — nothing enforces agreement, and the representations differ (app uses region *codes* like `SV1`; this list uses display city names). If the app's selectable-region set changes, this article will silently drift again — the same failure mode as ticket 546048.

**Ideal fix lives in dnsimple-app** (a reminder comment next to `Region::ALL`), but we don't have write access there. Options to route separately: open an issue on dnsimple-app, or add a CI tripwire here that diffs against the authoritative set. Flagging so a maintainer with app access can decide.

## Checklist

- [x] Branch created from `main` (confirm it's up to date with origin before merge)
- [x] Frontmatter has `title`, `excerpt`, `meta`, and `categories`
- [x] Opening paragraph directly answers the article's core question (SEO/GEO)
- [x] Cross-links added on first mention of related concepts
- [x] Existing articles updated to link back (if applicable)
- [x] Callouts use correct types (none added)

## Review

- [x] Second expert tagged for feature/product knowledge
- [x] Customer Success tagged (required for substantial updates)


lib/default.rb
```diff
@@ -37,3 +37,17 @@ def as_markdown(str)
   'Tokyo, Japan',
   'Warsaw, Poland'
 ].freeze
+
+# Regions a record can be pinned to with the regional records feature. This is a
+# deliberately smaller set than POPS: it must mirror Dnsimple::Region::ALL in
```
stephanoodle:
```suggestion
```

lib/default.rb
```diff
@@ -37,3 +37,16 @@ def as_markdown(str)
   'Tokyo, Japan',
   'Warsaw, Poland'
 ].freeze
+
+# Regions a record can be pinned to with the regional records feature. This is a
```
stephanoodle:
```suggestion
# Regions a record can be pinned to with the regional records feature.
```

san983 (approved):

PR #2012, opened 2026-07-07
Add Account new Explanation articles

Pemacaa:
## Summary

Adds three new Account **Explanation** articles and lists them in `categories/account.yaml` under Account and security concepts.

Closes dnsimple/dnsimple-customer-success#216

Part of dnsimple/dnsimple-customer-success#212

## New articles

| Article | File |
|---------|------|
| What is SSO at DNSimple | `content/articles/what-is-sso-at-dnsimple.md` |
| Users, Accounts, and Members at DNSimple | `content/articles/users-accounts-and-members-at-dnsimple.md` |
| What is Domain Access Control? | `content/articles/what-is-domain-access-control.md` |

## Modified

- `categories/account.yaml` — three new Explanation listings (titles match frontmatter exactly)

## Not in this PR (follow-up)

- Trim `domain-access-control.md` to How-to only (DAC split)
- Backlinks and anchor updates in existing articles

## Test plan

- [x] All three articles render with correct H1, TOC, and anchors
- [x] Category page lists new articles under Account and security concepts with working links
- [x] YAML titles match frontmatter `title` verbatim
- [x] No account email terminology in new articles


stephanoodle:
## Why the DAC split + backlinks are in this PR (not a follow-up)

The original plan listed the `domain-access-control.md` trim and the backlink/anchor updates as follow-up work. We folded them into this PR instead, because they are **atomically coupled** to the new Explanation articles and cannot be safely separated:

1. **The three pieces depend on each other.**
   - The new Explanation articles *define* the section anchors (`#full-access`, `#limited-access`, `#domain-manager`, `#zone-operator`).
   - Trimming the DAC how-to *removes* those same anchors from their old location.
   - Every inbound link across the site (contact-management, name-server-sets, the integrated-DNS guides, the IdP guides, etc.) has to be repointed to the new location.

   The moment the how-to is trimmed, all of those inbound links break unless the repoints land in the same change. So the trim and the repoints can never live in different PRs.

2. **Splitting would either break `main` or ship the exact overlap we're removing.**
   - If we merged the Explanation articles first and deferred the trim, the live site would temporarily have the how-to's explanatory content *and* the new Explanation article saying the same thing — the duplicate content this work exists to eliminate.
   - If we merged the repoints before the new articles existed, they'd point at anchors that don't exist yet.

3. **The only clean seam is "before vs. after the new articles," and everything after has to travel together** — so one PR is the smallest change that leaves `main` link-consistent at every merge point.

Verification: an anchor-resolution pass over `content/articles/` confirms all moved DAC role anchors now resolve to the Explanation article, with no stale `domain-access-control/#…` references left behind.

Unrelated pre-existing broken anchors found during that pass were split out into a separate hygiene PR (#2013) to keep this one focused on the Account/DAC/SSO scope.

Pemacaa:
Omg thank you so much for this! TIL

PR #2026, opened 2026-07-17
Isolate the support widget in a shadow root

weppos:
The support widget is injected into other DNSimple sites, and its CSS was appended to their `<head>` by `vite-plugin-css-injected-by-js`. Every bare selector the widget ships then applied to the host page. On dnsimple.com the app's own `.callout` component picked up the widget's unrelated `.callout` rules and rendered with a stray black left border and a grey wash.

This isolates the widget in a **shadow root** so the boundary is structural rather than a naming convention the next rule can forget. Nothing the widget ships reaches the host page, and nothing the host ships reaches the widget, permanently.

**No change to the support site itself**, its own callouts and pages render identically.

### DIFF
--- a/_widget/src/components/app/component.vue
+++ b/_widget/src/components/app/component.vue
@@ -24,11 +24,6 @@ import Articles from '../articles/component.vue';
 import Prompt from '../prompt/component.vue';
 import Search from './search.js';
 
-import "./variables.scss";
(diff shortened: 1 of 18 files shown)


dallasread (approved):
QA's 💯. Thanks for the fix.

PR #2047, opened 2026-07-24
Refresh dnsimple.com WHOIS sample in domain-resolution-issues

juankuquintana:
Belongs to https://github.com/dnsimple/dnsimple-business/issues/2876

This PR regenerates the stale 2013-era WHOIS block against the current thin-registry response, so the article no longer implies dnsimple.com is with Enom or has already expired.


Pemacaa (approved):

Pemacaa:
Thank you!

PR #2052, opened 2026-07-24
Update Contacts Explanation articles for Diataxis revamp

stephanoodle:
icann-60-day-lock-registrant-change.md: the opening sentence was
ungrammatical and ended in a colon introducing a quotation that sits
further down the page. Rewritten as a direct answer. Fixes the H1 to
match the frontmatter title.

Adds a 'When the lock applies' section. The article never said what
triggers the lock, which is the most common misunderstanding on this
topic. Verified against DomainRegistrantChangePrepareInteractor: the
lock is flagged only when the suffix is an ICANN suffix and the change
is a registry owner change, so editing a phone number or address on the
existing registrant does not trigger it. Also documents that the app
warns before you confirm, since customers arrive here from that warning.

icann-domain-validation.md: adds H2 anchors and a section contrasting
ICANN registrant validation with NIS2 contact email verification. Only
one of the two suspends a domain, and customers receive both.

what-is-domain-trustee.md: adds a section on trustee and extended
attributes, and a return link to Changing Domain Contacts.

Part of Phase 2 of the Category Documentation Revamp Plan (dnsimple/dnsimple-customer-success#200).

Closes dnsimple/dnsimple-customer-success#226

### DIFF
--- a/content/articles/what-is-domain-trustee.md
+++ b/content/articles/what-is-domain-trustee.md
@@ -20,13 +20,13 @@ Some top-level domains (TLDs) require local presence, a designated registrant ty
 
 ## What Is a Domain Trustee? {#what-is-a-domain-trustee}
 
-A **domain trustee** is the party that meets the registry's eligibility role on paper—often a local partner or designated registrant—when rules say the domain cannot be held under your personal or company details alone. **You remain the beneficiary**: you keep operational control through DNSimple (DNS, renewal, transfers, and account billing). The trustee fulfills the registry's legal or administrative requirement so the registration or transfer can complete.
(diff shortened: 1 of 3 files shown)


content/articles/what-is-domain-trustee.md
```diff
@@ -50,6 +50,10 @@ When trustee is optional and you add it, DNSimple lists trustee as its own invoi
 
 Yes. If you register or transfer domains using the [DNSimple API](/articles/api-documentation/), the same optional-or-required trustee rules apply as in the dashboard. Use the [developer documentation](https://developer.dnsimple.com/) for the exact requests and responses for your integration.
 
+## Where to manage trustee settings {#manage}
+
+Trustee applies per domain name. When you [change a domain contact](/articles/changing-domain-contact/) on a domain that uses trustee, the extended attributes shown are limited to those that still apply for that trustee configuration.
```
Pemacaa: What does this mean?

Pemacaa (approved):

content/articles/what-is-domain-trustee.md
stephanoodle: You're right that that doesn't make sense. 

I rewrote some of this to hopefully make more sense. 

Let me know what you think.

PR #2060, opened 2026-07-24
Add What Are Connectors? explanation article

stephanoodle:
Flagged as a likely gap in the Phase 2 plan. Both connector articles
open on an OAuth flow without saying what a connector is, what the
authorization grants, how it differs from an integration or a service,
or that connections are monitored after setup.

Content verified against dnsimple-app: connectors are Addon records
backed by an OAuth application, which is why setup authorizes both
sides; a connection is a PlatformConnection scoped to an account and a
resource, carrying title, metadata, messages, and status history;
deleting one removes the connection and its status, message, and
metadata records. MarketingPages::CONNECTORS confirms Heroku and
Netlify are the only two.

The deletion note deliberately does not claim what happens to the DNS
records the connector created, since that is not determinable from the
deletion path alone. It tells the reader to check instead.

The connector row of the comparison table is verified. The integration
and service rows are the working definitions for the Phase 5 and Phase 6
boundaries and should be confirmed with an SME before this merges.

Adds the Explanation section to connectors.yaml. Note the YAML assumes
the Connectors Diataxis restructure has merged; land that PR first.

Part of Phase 2 of the Category Documentation Revamp Plan (dnsimple/dnsimple-customer-success#200).

> [!IMPORTANT]
> `categories/connectors.yaml` here assumes the Connectors Diataxis restructure has merged. Land that PR first, then `git merge main` into this branch.

## Needs SME confirmation

Please confirm the connector / integration / one-click service comparison table before merge. The **connector** row is verified against `dnsimple-app`. The integration and service rows are the working Phase 5 and Phase 6 boundary definitions and should be checked by product or engineering.

Also confirm what happens to the DNS records a connector created when a connection is deleted. The article deliberately does not assert this and tells the reader to check instead.

Closes dnsimple/dnsimple-customer-success#234

PR #2065, opened 2026-07-24
Add Connectors at DNSimple pillar page

stephanoodle:
Connectors' main navigation problem is not size but ambiguity. A
customer looking for how to point a domain at Heroku may land in
Connectors, Integrations, or Services, and nothing told them which one
they wanted. The orientation section is the reason this page exists.

The opening states that Heroku and Netlify are the only connectors, so
readers stop looking for others.

The comparison table matches What Are Connectors?: connectors vs one-click
services, with integrations pointed to as the path for other providers.
The connector row is verified against `dnsimple-app`. The one-click service
row is the working Phase 6 boundary definition and should be confirmed with
an SME. Whatever ships here needs to stay aligned with the Integrations
revamp in Phase 5.

Adds the Getting started section to `connectors.yaml`, and lists the pillar
under Connectors in `integrations.yaml` so the Integrations frontmatter
category does not drop it into Other.

Part of Phase 2 of the Category Documentation Revamp Plan (dnsimple/dnsimple-customer-success#200).

## Needs SME confirmation

Please confirm the one-click service row in the comparison table before merge.
The **connector** row is verified against `dnsimple-app`. The service row is
the working Phase 6 boundary definition and should be checked by product or
engineering.

Closes dnsimple/dnsimple-customer-success#239


content/articles/connectors-at-dnsimple.md
```diff
@@ -0,0 +1,51 @@
...
+- [What Are Connectors?](/articles/what-are-connectors/) - What a connection stores, what authorizing a connector grants, how monitoring works, and what deleting a connection does.
+
+## Choosing between connectors, integrations, and services {#choosing}
+
+These three overlap, and which one you want depends on what you are connecting to.
+
+| | Use when | Where to look |
+|---|---|---|
+| **Connector** | You use Heroku or Netlify and want DNSimple to configure and monitor the records | This category |
+| **Integration** | Your provider is not Heroku or Netlify, or you need a provider-specific setup path | [Integrations](/categories/integrations/) |
```
lokst: IMO, `Integration` might be a little confusing so I'd consider removing this mention (see: https://github.com/dnsimple/dnsimple-support/pull/2060/changes#r3673495548)

lokst (approved):

content/articles/connectors-at-dnsimple.md
stephanoodle:
```suggestion
```

Pemacaa (approved):

PR #2080, opened 2026-08-14
Point row-action steps at the contextual menu

stephanoodle:
Closes dnsimple/dnsimple-customer-success#241 (copy half).

dnsimple/dnsimple-app#36335 moved per-row actions behind a single trailing `...` menu. Articles that told the reader to click an edit, trash, or note icon on a row now point at the menu and name the item to select. A follow-up pass covers listings whose steps already named the button (Edit, Remove, Renew, Revoke, Disconnect) without calling it an icon.

Labels were taken from the shipped views rather than guessed. The audit is written up in [this comment](https://github.com/dnsimple/dnsimple-customer-success/issues/241#issuecomment-5297204864), including the eight files the issue's initial grep missed and the two it listed that turned out to be unaffected.

## What changed

**Record editor family**

- `record-editor.md`, `record-editor-integrated-zones.md`
- `manage-a-record.md`, `manage-aaaa-record.md`, `manage-cname-record.md`, `manage-caa-record.md`, `manage-url-record.md`, `manage-tlsa-record.md`, `manage-service-binding-records.md`
- `change-ttl.md`, `managing-email-dns-records.md`, `managing-mx-records-when-changing-email-providers.md`, `setting-up-null-mx-records.md`

**Record notes**

- `create-record-note.md` - the speech-bubble icon is now the **Add note** item, which reads **View note** when the record already has a note
- `create-record-deletion-note.md` - the deletion note is reached through **Delete** in the menu

**Other listings**

- `contact-management.md` - **Edit** is in the menu; **Delete** only appears when the contact is deletable, matching the old icon behavior
- `changing-domain-contact.md` - **Edit** on the contacts list
- `add-and-remove-email-forwards.md`, `email-forwarding-management.md`
- `api-access-token.md` - **Edit** appears for scoped tokens, **Delete** for all
- `name-server-sets.md`, `name-servers-interface-reference.md`
- `webhooks.md` - covers both **Delete** and **Clear Suppression**
- `services.md` - **Remove**
- `account-users.md`, `managing-seats.md`, `domain-access-control.md` - **Resend**, **Revoke**, **Remove**, **Manage Access**
- `google-identity-provider.md`, `okta-identity-provider.md`, `entra-identity-provider.md` - **Revoke** on SSO workspaces and user identities (Okta's identity step previously said **Remove**)
- `integrated-dns-provider-zones.md` - **Remove** and **Delete**
[...]

PR #2093, opened 2026-08-19
Add weekly screenshot refresh action

dallasread:
## Summary

Adds an automated system that keeps the screenshots in `content/files/` in sync with the live app. Once a week (and on demand), a GitHub Action logs into the app, re-captures every screenshot that has a capture annotation, and opens a single reviewable PR containing only the images that actually changed.

## How it works

**1. Articles declare what to capture.** The capture spec is an HTML comment next to the image it refreshes, invisible in the rendered article:

```markdown
![screenshot of the API & Access page](/files/access-tokens.png)
<!-- capture: /a/{account}/account/api_tokens -->
```

- The path is app-relative. `{account}` is auto-detected after login from the app's own `/a/<id>/` links.
- An optional `#fragment` clips the shot to one element (element id first, then a `data-screenshot` attribute, then a raw CSS selector). No fragment = full 1280x800 viewport at 2x (retina), matching existing screenshots.
- Images without an annotation are never touched. Coverage grows one annotation at a time; this PR seeds the first one (`access-tokens.png` in the API Access Token article).

**2. The runner captures and filters.** `_screenshots/capture.js` logs into https://sandbox.dnsimple.com with a dedicated account, resolves the app origin from wherever login lands (sandbox redirects to app.sandbox.dnsimple.com), hides sandbox-only chrome (the red ribbon, the strip under the navbar, the floating help widget), captures each spec, and compresses with sharp (palette quantization, no system packages). A capture that hits a non-2xx response or a login redirect fails loudly instead of saving a bad image.

**3. Only real changes ship.** Old and new images are pixel-diffed; a file is replaced only when more than 0.1% of pixels differ beyond an anti-aliasing threshold. An unchanged app produces no PR at all.

**4. One refresh PR, ever.** Changes land on the fixed `refresh-app-screenshots` branch. Reruns force-update the same open PR (recomputed against current main each time, so it never stacks or conflicts), and the PR auto-closes if the drift disappears.

## Triggers

| Trigger | Behavior |
|---|---|
| Cron, Mondays 09:00 UTC | Opens or updates the refresh PR |
| Run workflow (no input) | Same as the cron |
| Run workflow with a PR number | Captures on that PR's branch and commits changed images back to it |
| `/refresh-screenshots` commented on a PR | Same as above; org members and collaborators only |

The PR-targeted triggers cover re-capturing an image mid-review and content PRs that add annotations. Fork PRs are not supported (the workflow token cannot push to forks).

## Verified

- Unit tests (`npm test`) cover the annotation parser, spec conflict detection, placeholder resolution, and the pixel-diff decision.
- `node test/smoke.js` runs the whole pipeline against a local fake app with zero credentials.
- A full CI run against the real sandbox went green end to end and opened #2094 as proof (test artifact, safe to close).
- The committed `access-tokens.png` is a real capture from this pipeline.

## Setup status

- [x] `SCREENSHOT_EMAIL` / `SCREENSHOT_PASSWORD` repository secrets (sandbox account, 2FA off)
- [ ] PR-creation authorization: pending dnsimple/infra-github#351, which lets the default `GITHUB_TOKEN` open PRs. Until it merges and applies, the weekly run fails at its final step. Alternative: a `SCREENSHOT_PR_TOKEN` secret (PAT with contents + pull-requests write) - it takes precedence when present, and PRs it opens also trigger CI, which default-token PRs cannot.

## Known limitations

- The sandbox account's content (plan, tokens, domains) shows in the screenshots; a follow-up will seed and reset it to a fixed state before each run so captures are deterministic.
- The first CI capture of any locally-captured baseline reports a change (macOS vs Linux font rendering); captures are stable CI-to-CI afterwards.
- `.markdownlint.yaml` is included because the pre-commit hook and markdownlint CI expect it; it disables the four default rules that conflict with existing article conventions (line length, inline HTML, single-title, heading increment).


stephanoodle:
Claude and I went through this with an eye to annotating the rest of the library, and ran the pipeline locally against the branch (`npm test` passes 9/9, including the smoke test). A few findings, one of which is a straightforward bug.

### The filename pattern drops 13 existing screenshots silently

`IMAGE_RE` in `_screenshots/lib.js` only matches lowercase:

```js
const IMAGE_RE = /!\[[^\]]*\]\(\/files\/([a-z0-9-]+\.png)\)/g;
```

13 of the images referenced from articles have capitals or underscores: the nine `DNSSEC-*-reference.png` files, plus `edit_notifications.png`, `notification_card.png`, `reissue_commercial_cert_with_rsa.png`, and `getting-started-transfer-WHOIS-domain.png`. I checked against the real `extractSpecs` that a capture comment on any of them produces no spec at all, with no warning and nothing in `--list`. Widening to `[A-Za-z0-9_-]+` would cover them.

### The first capture of any image always reports a change

`pngsDiffer` returns true on a size mismatch before it compares any pixels. The existing library has 475 distinct pixel sizes across 509 images, and none of them are 2560x1600, so every screenshot we annotate comes back "changed" on its first run whether or not the app changed. `access-tokens.png` in this PR shows it: 2232x1120 to 2560x1600.

Flagging it mainly so nobody reads that first run as a drift report. I actually think the re-shoot is worth having: uniform screenshot sizes across the library would make the documentation better on its own. What we'd need is a plan for how to tackle ~500 images, mostly how they get reviewed and in what size batches. Drift detection proper starts from the second run onward.

### The coverage ceiling is lower than it looks

`capture()` is `goto` plus an optional element clip, so anything needing a click can't be captured. A large share of our screenshots are modals, confirmations and post-action states: `add-zone-modal.png`, `a-record-created.png`, `secondary-dns-confirmation-message.png`. Another big chunk are crops of buttons, tabs and links that would each need a `data-screenshot` hook in dnsimple-app.

`resolvePath` also only knows `{account}` and `{domain}`, so pages addressed by a record, certificate, contact or invoice id can't be expressed as a spec at all.

Rough split of the 509: ~218 element crops, ~113 interaction states, ~57 third-party UI, diagrams or email screenshots, and only about 20-25 that are whole pages reachable by a plain GET with the placeholders available today.

Keeping the docs current is the goal here, not refreshing a handful of index pages. So some way to describe a short interaction before capture is the highest-value thing to add next: it's what decides whether this covers most of the library or a corner of it.

### Small thing

`activity-tracking-dashboard.png` and `user-activity-log.png` are different crops of the same page, so annotating both to `/a/{account}/account/activity` would produce two identical files. `mergeSpecs` checks per file, so it wouldn't flag it.

None of this blocks merging.

_screenshots/capture.js
```diff
@@ -0,0 +1,197 @@
...
+  const tmpFile = path.join(tmpDir, spec.file);
+  if (spec.fragment) {
+    let target = null;
+    for (const selector of fragmentToSelectors(spec.fragment)) {
+      const locator = page.locator(selector).first();
+      if (await locator.count()) { target = locator; break; }
+    }
+    if (!target) throw new Error(`no element matches #${spec.fragment} on ${url}`);
+    await target.scrollIntoViewIfNeeded();
+    await target.screenshot({ path: tmpFile });
```
stephanoodle, line 111: A `<dialog>` is in the page but closed, so it has no box to photograph and `element is not visible` is all you get. Opening it in place is enough, since `DialogComponent` renders every dialog into the page with a stable id and `data-dialog-open` triggers.

Shoot `[data-dialog-panel]` rather than the dialog element, which spans the viewport once modal.

The `scrollIntoViewIfNeeded` call also has to go for this path. A modal panel is already centred, and on `#unsubscribe-dialog` that call hung for the full 30s and failed the run. The two dialogs I captured only came out once it was skipped.

```suggestion
    // A <dialog> is in the page but closed, so it has no box to photograph.
    // Open it in place, the way the app's own data-dialog-open trigger would,
    // then shoot the panel rather than the dialog element, which spans the
    // viewport once modal. A modal panel is already centred, so scrolling it
    // into view is unnecessary and in practice hangs.
    if (await target.evaluate((el) => el.tagName === 'DIALOG')) {
      await target.evaluate((el) => { if (!el.open) el.showModal(); });
      await page.waitForTimeout(300);
      const panel = target.locator('[data-dialog-panel]').first();
      if (await panel.count()) target = panel;
    } else
      await target.scrollIntoViewIfNeeded();

    await target.screenshot({ path: tmpFile });
```

Verified against `#domain-list-labels-modal` and `#external-provider-route53-dialog`; both produced clean panel-only images at 1024px wide.

_screenshots/capture.js
```diff
@@ -0,0 +1,197 @@
...
+  }
+  const email = process.env.SCREENSHOT_EMAIL || CONFIG.email;
+  const password = process.env.SCREENSHOT_PASSWORD || CONFIG.password;
+  if (!email) throw new Error('SCREENSHOT_EMAIL must be set');
+  if (!password) throw new Error('SCREENSHOT_PASSWORD must be set');
+
+  const baseUrl = (process.env.SCREENSHOT_BASE_URL || CONFIG.baseUrl || 'https://sandbox.dnsimple.com').replace(/\/$/, '');
+  const vars = {
+    account: process.env.SCREENSHOT_ACCOUNT || CONFIG.account,
+    domain: process.env.SCREENSHOT_DOMAIN || CONFIG.domain,
```
stephanoodle, line 137: One `{domain}` cannot cover the dialog batches. They need a DNSSEC-signed domain, one with vanity name servers, one with a linked integrated provider and one with records, sometimes in the same article.

Naming them `{domain_dnssec}` keeps the existing `\{(\w+)\}` regex in `resolvePath` working, so this is the only file that changes.

```suggestion
    domain: process.env.SCREENSHOT_DOMAIN || CONFIG.domain,
    // Domains in a particular state, so one spec can ask for the DNSSEC-signed
    // domain and another for the one with vanity name servers:
    // {domain_dnssec} resolves from SCREENSHOT_DOMAIN_DNSSEC, and so on. The
    // dialog batches need several at once, which a single {domain} cannot cover.
    ...Object.fromEntries(Object.entries(process.env)
      .filter(([name]) => name.startsWith('SCREENSHOT_DOMAIN_'))
      .map(([name, value]) => [`domain_${name.slice(18).toLowerCase()}`, value])),
```

Worth pairing with a clearer failure message in `resolvePath`, which currently names `SCREENSHOT_DOMAIN_DNSSEC` correctly by accident.

_screenshots/lib.js
```diff
@@ -0,0 +1,96 @@
...
+  const specs = [];
+  const lines = markdown.split('\n');
+  for (let i = 0; i < lines.length; i += 1) 
+    for (const match of lines[i].matchAll(IMAGE_RE)) {
+      const file = match[1];
+      const rest = lines[i].slice(match.index + match[0].length);
+      let comment = rest.match(CAPTURE_RE);
+      if (!comment) {
+        let j = i + 1;
+        while (j < lines.length && lines[j].trim() === '') j += 1;
```
stephanoodle, line 22: Unrelated to the dialogs, but it cost me a capture and is cheap to guard.

A `capture:` comment trailing an image line also binds to the *previous* image when two images sit on consecutive lines, because the scan for the next non-blank line does not stop at another image. Annotating `user-activity-log.png` this way silently claimed `view-activity-log.png`, a crop of one link, and would have overwritten it with a full page.

`node capture.js --list` catches it, so the cheapest fix may be documenting that check in the README rather than changing the parser. If you would rather it be structural, stop the forward scan when the line already contains an image reference.

stephanoodle (commented):
Ran this harness against the sandbox account to try to fill in the dialog batches (#2081, #2082, #2083) and the row-action screenshots (#2080). Two suggestions below are things I hit directly, and one is a limit those batches run into immediately.

**What worked.** With the first suggestion applied, `enter-labels.png` and `account-external-provider-link-credentials.png` captured cleanly and are now committed to #2083 and #2081. Dialogs turn out to be easy targets: `DialogComponent` renders them into the page with stable ids, so nothing has to be clicked, only opened.

**What stopped everything else.** The screenshot account is on the **Solo plan with zero domains**. Every remaining dialog in those batches lives on a domain or zone page, and `/a/2944/zones/new` returns 403 on this plan. The row menus in #2080 need populated tables; a crawl of 32 pages found exactly one row menu on the whole account. Per our own [sandbox article](https://support.dnsimple.com/articles/sandbox/), every action in the sandbox is free and plans can be subscribed with Stripe test cards, so this is seeding rather than a cost question.

None of the three suggestions are needed for the account pages already captured in #2094 and #2113. They are what the dialog batches need.

PR #2096, opened 2026-08-19
Align article H1s with frontmatter titles and update category entries

stephanoodle:
`article-structure.mdc` says the H1 matches the frontmatter `title`. 155 articles disagree. This fixes the 45 in the Diátaxis-treated categories (DNS, DNSSEC, Domains and Transfers, SSL Certificates, Name Servers).

## Why it matters beyond tidiness

The two fields feed different things:

| Field | Consumed by |
|---|---|
| frontmatter `title` | `<title>` tag → browser tab and the Google result link · `search.json` · category listings · category YAML matching |
| H1 | the rendered page heading only |

So on `caa-record`, the tab, the search result, the search index, and the category listing all said *What Is a CAA Record?* while the page heading said *CAA Records*. Readers clicked one thing and landed on another.

## Direction, per article

Not one rule applied 45 times — the side carrying more information wins, which isn't always the longer one.

- **33 articles: H1 rewritten** to match frontmatter. Includes 15 question-form conversions (`A Records` → `What Is an A Record?`), which also helps the question-shaped queries that retrieval currently handles worst
- **10 articles: frontmatter rewritten** to match the H1, where the H1 genuinely held more — `Zone NS Records` → `Updating Zone NS Records for a Hosted Domain`, `DNSSEC at DNSimple` → `DNS Security Extensions (DNSSEC) at DNSimple`
- **1 article: both rewritten** — `delegating-dnsimple-registered` was awkward either way, now `Delegate a DNSimple-Registered Domain to DNSimple Name Servers`
- **1 article skipped** — `what-is-dig`. The H1 is ``What Is `dig`?`` with code formatting; the title field is interpolated raw into `<title>` and listing links, so backticks would render literally there. The fields differ only by markup that one of them cannot carry, and that is fine

## Category YAML

`lib/sub_categories.rb` matches articles to category entries by **exact title string** — a rename without a paired YAML update silently drops the article into "Other." 16 entries updated across 8 files; several titles appear in two or three YAMLs.

Also folded in: **3 pre-existing orphans** whose titles were in no category YAML at all — `integrated-domain-providers`, `what-is-domain-access-control`, `what-is-sso-at-dnsimple`.

## Verification

Simulated the exact-string matching across the whole corpus, before and after:

- **Orphans: 3 → 0 corpus-wide**
- **Stale entries: 21 → 21, none introduced**
- All 21 category YAMLs parse
- Title mismatches in the five categories: 45 → 1 (the deliberate `what-is-dig` exception)

⚠️ `bundle exec rake` has **not** been run against this branch — worth doing before review.

## Not included

The 57 `.XXX Domains` / `.XXX Domain Names` TLD articles. That's one decision repeated 57 times rather than 57 decisions, and it needs a product-voice call since the app's own nav label is *Domain Names*.

### DIFF
--- a/categories/dns.yaml
+++ b/categories/dns.yaml
@@ -56,7 +56,7 @@ Explanation:
     - How Does Zone Verification Work?
     - What Is Zone Verification and Why Does It Matter?
     - How Long Does a New DNS Record Take to Resolve?
-    - How DNS Caching and TTL Affect Record Changes
+    - How DNS Caching and TTL Affect Delegation and Record Changes
     - Recursive vs Authoritative DNS Resolvers
     - Protection Against DDoS Attacks
     - Why DNSSEC and Secondary DNS May Not Work Together
@@ -91,7 +91,7 @@ How to:
     - Activate and Deactivate a DNS Zone
     - Export Zone File
     - Import Zone File
-    - Zone NS Records
+    - Updating Zone NS Records for a Hosted Domain
     - Check Resolution Status
     - Check DNS Cache
     - How to Use dig
@@ -116,7 +116,7 @@ How to:
     - Refresh and Import Integrated Zones
     - Sync Integrated Zone Records Between DNS Providers
     - How to Use CoreDNS as an Integrated DNS Provider
-    - How to Use Route 53 as an Integrated DNS Provider
+    - How to Use Amazon Route 53 as an Integrated DNS Provider
     - Auto-Import DNS Records
   DNSSEC Management:
     - Add and Remove DS Records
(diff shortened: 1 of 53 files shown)


Pemacaa (approved):

Pemacaa:
This looks good! 

One note for a follow-up PR, some links in other articles still use the old titles as the clickable text. The URLs are fine, but the label doesn’t match the new page title. For example:

Links still say Zone NS Records, but the page is now Updating Zone NS Records for a Hosted Domain
Links still say DNSSEC at DNSimple, but the page is now DNS Security Extensions (DNSSEC) at DNSimple
Same pattern for the two delegation articles, Regulated TLDs, and the Route 53 guide.

We can take that later :)

PR #2106, opened 2026-08-27
Fix link issues surfaced by Ahrefs report

stephanoodle:
An Ahrefs crawl flagged `/articles/how-to-determine-certificate-authority/`
as a 502. That turned out to be a transient error from our host and the page
is healthy, but checking how the article is linked surfaced two real issues.

## Stale link labels

Three links to the article used its H1 ("How Do I Determine the Certificate
Authority That Signed My SSL Certificate?") as their clickable text instead of
its title ("How to Find Your SSL Certificate Authority"). This is a follow-up
to #2096 and the later label pass; these three were missed.

URLs were already correct. Label text only.

- `content/articles/what-is-certificate-authority.md`
- `content/articles/what-is-ssl-root-certificate.md`
- `content/articles/ssl-certificate-authorities.md`

The generated link on `/categories/ssl-certificates/` already used the correct
title, since it is built from frontmatter. Only the hand-written ones drifted.

## Wrong redirect target

`/articles/how-to-apply-github-student-pack` redirected to the certificate
authority article. Its three sibling Student Pack entries all point to
`/articles/github-pages/`, which is where this one belongs. Anyone following
an old Student Pack link landed on SSL certificate instructions.

Confirmed live before the fix: the old URL returned a 301 to the SSL article.

## Note on scope

This mixes a content change with an infrastructure change, which the repo's
git-workflow rule normally separates. Kept together deliberately because both
came out of the same report and each is a one-line fix. Happy to split if
preferred.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -51,9 +51,9 @@
 /articles/how-can-i-issue-a-new-transfer-order-when-a-transfer-is-denied  /articles/how-to-issue-new-transfer-when-transfer-denied/
 /articles/how-can-i-issue-a-new-transfer-order-when-a-transfer-is-denied/  /articles/how-to-issue-new-transfer-when-transfer-denied/
 /articles/how-can-i-issue-a-new-transfer-order-when-a-transfer-is-denied/index.html  /articles/how-to-issue-new-transfer-when-transfer-denied/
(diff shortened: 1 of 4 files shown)


Pemacaa (approved):
Thank you for fixing this!

PR #2150, opened 2026-09-16
Pin authorization code and account move searches

stephanoodle:
Adds exact-match pins to `_widget/src/components/app/rigged-results.yml` for short searches that the site search ranks poorly, following #2149.

Search runs Fuse.js fuzzy matching, which is loose on one- and two-word queries. After #2149, several of the names customers use still land on unrelated articles:

- `auth code`, `authcode` - Discount Codes first
- `epp`, `auth info` - the right article not in the top 7
- `move domain` - a TLD article first
- `domain push`, `account identifier` - the right article not in the top 7

**Authorization code** - pinned to `domain-transfer-code-reference`, `transferring-domain-away`, `troubleshooting-domain-transfer-issues`:

`authorization code`, `authorisation code`, `auth code`, `authcode`, `auth info`, `epp code`, `epp`, `transfer code`

These are the names customers use in tickets. Across 164 tickets since January 2025 that mention the code: authorization code 61, auth code 48, EPP code 32, transfer code 26. `transfer code` and `epp code` both dropped a place after the reference article was retitled, so they are pinned too.

**Account moves** - pinned to `transferring-domain-between-accounts`, `getting-started-clickfunnels`:

`move domain`, `domain push`, `push identifier`, `account identifier`

`push identifier` covers the Domain Push Identifier name that ClickFunnels still asks customers for.

The existing `transfer` pin is unchanged. All pinned articles exist.


dallasread (approved):
😎. Amazing – so happy to see this!

Pemacaa (approved):

PR #2152, opened 2026-09-17
Update ALIAS failure behavior and negative caching time

stephanoodle:
Updates two articles to match recent name server releases.

## ALIAS records (`alias-record.md`)

[erldnsimple 16.11.0](https://github.com/dnsimple/erldnsimple/releases/tag/16.11.0) ([#893](https://github.com/dnsimple/erldnsimple/pull/893)) changed what happens when an ALIAS lookup fails. The article said we return an empty result, which is the old behavior. It now covers:

- Each attempt has 500 ms, with 2 attempts total (the old text said "a few times")
- A failed lookup returns `SERVFAIL`, so resolvers retry instead of caching an empty answer
- A target that does not exist returns an empty answer instead of `NXDOMAIN`, so other records on the name (MX, TXT) keep working

Checked against `lib/erldnsimple/alias.ex` and the failure table in `docs/distribution-semantics.md`.

## Negative caching (`how-long-does-it-take-for-a-new-record-to-resolve-for-my-domain.md`)

[erldnsimple 16.12.0](https://github.com/dnsimple/erldnsimple/releases/tag/16.12.0) ([erldns #370](https://github.com/dnsimple/erldns/pull/370)) fixed negative answers being held for the full SOA TTL (1 hour) instead of the SOA minimum. The article said negative results can take "up to an hour" to clear. It now says the SOA minimum TTL controls this, DNSimple zones default to 5 minutes, and zones with a higher minimum wait longer.

Tested live on September 17, 2026:

- `dnsimple.com` (SOA minimum 300): our name servers and 1.1.1.1, 8.8.8.8 and 9.9.9.9 all serve the negative SOA at 300
- A customer zone with SOA minimum 3600: served at 3600, so the default is not universal
- New zone default of 300 confirmed in `DEFAULT_MINIMUM` in the app (`lib/dnsimple/rr/r_data/soa.rb`)

"About 10 minutes" is edge cache (5) plus resolver cache (5). The resolver part matches the live test; the edge part comes from the #370 deployment verification.

## Not changed

[erldnsimple 16.13.0](https://github.com/dnsimple/erldnsimple/releases/tag/16.13.0) (DNSSEC at delegation points) needs no doc change. DS is a system-only record type, so customers still cannot publish DS records for a delegated subdomain, which is what the DNSSEC articles say.

The ALIAS article still says the handler uses "a local PowerDNS resolver". The code only shows a local resolver at 127.0.0.1, so a reviewer who knows the setup should confirm that line.


NelsonVides (approved):
Ooooh thank you so much, had no idea about these articles, good catch! ❤️

Pemacaa (approved):

stephanoodle:
> Ooooh thank you so much, had no idea about these articles

No problem! This is why posting as you did in the channel is the best <3

PR #2153, opened 2026-09-17
Check link anchors in CI

stephanoodle:
The link check validates that a linked page exists. It does not check whether the `#anchor` is actually in that page, because lychee ignores URL fragments unless asked. So renaming a heading sends every inbound link to the top of the article instead of to the step it promised, and CI stays green.

`content/` has 340 links pointing at a specific heading. The most-linked single anchor, `/articles/services/#removing-services`, is used 37 times.

## The change

Two flags on the existing check, which stays offline against the compiled output:

- `--include-fragments` — resolve the anchor, not only the page.
- `--index-files index.html` — **required with it.** Our links point at directories (`/articles/foo/`), and lychee cannot read ids out of a directory. With the first flag alone, all 373 anchored links fail as "Cannot find fragment," which is likely why this was never turned on.

## What it caught: 26 broken anchors in 16 files

Two kinds, both fixed here by correcting the links. No heading, id, or URL changes.

**20 links with a slash after the anchor** — `/articles/foo/#anchor/`. The trailing slash makes the fragment `anchor/`, which matches nothing, so the reader lands at the top of the page. Worst offenders were `/articles/integrated-dns-provider-zones/#removing-integrated-zones-from-dnsimple/` and `#deleting-a-zone-from-an-integrated-dns-provider/`, four uses each.

**6 links using the auto-generated slug where the heading declares a shorter id:**

| Link said | Heading has |
|---|---|
| `#updating-a-domain-contact` | `{#updating}` |
| `#removing-members-from-an-account` | `{#removing-members}` |
| `#transfer` / `#register` (in `domains-uk`) | `{#transferring}` / `{#registering}` |
| `#sync-dns-records-from-route-53` | `{#syncing-dns-records-from-route-53}` |
| `#name-server-propagation-delay` | `{#check-name-server-propagation-delay}` |

That second group is the category worth having a check for: the link matches what the heading text would generate, so it looks right in review, and it has been wrong the whole time.

## Note for review

I tried to confirm up front that this would not turn the check red, and got it wrong: my pre-check treated both the generated slug and the explicit `{#custom-id}` as valid anchors, while kramdown only emits the explicit one. CI found the 26. All green now.


dallasread (approved):
Nice find.

PR #2156, opened 2026-09-18
Check the mechanical writing rules in CI

stephanoodle:
The rules in `.cursor/rules/article-writing.mdc` are enforced by hand today, after the fact, in cleanup commits. Some of them are mechanical and do not need a person: the em-dash and contraction sweeps, missing frontmatter, deprecated terms, callout types, link form. This adds a check for the ones a machine can decide.

Also the cleanest answer to output quality: a linter does not require anyone to accept feedback.

## What it is

`_test/writing_rules_test.rb`, in the existing `rake test:ruby` suite. No new gems, no new workflow, no compile step - it reads `content/articles/*.md` directly, the same way `images_test.rb` does.

Eight checks:

| Check | Rule |
|---|---|
| Frontmatter has `title`, `excerpt`, `meta`, `categories` | review checklist |
| Article ends with `## Have more questions?` | structural patterns |
| Callouts are only NOTE, TIP, WARNING | review checklist |
| Internal links are root-relative and end in a slash | cross-linking |
| TLD articles end in `Domains`, title and H1 | terminology |
| No deprecated terms (Standard certificate, Comodo) | terminology table |
| Preferred spellings (name server, subdomain, organization, auto-renewal, top-level domain) | preferred terms |
| No banned phrases (throat-clearing openers, filler, hedging adverbs) | anti-patterns |

Failures name the article and the phrase, and point back at the rule file.

## The 21 fixes that make it pass

19 articles: 3 `[!INFO]` callouts became `NOTE`, 4 "Top Level Domain" became "top-level domain", one "sub-domain" and one "Organisations" corrected, 2 uses of "essentially" dropped, and 10 one-word "nameserver" in prose and image alt text became "name server".

## Three exemptions, because the flat rule was wrong

Worth reviewing, since these are the judgement calls:

- **"formerly Comodo" is allowed.** Both remaining uses explain the Sectigo rebrand, which is what the terminology table asks for.
- **A quoted `"nameserver"` is allowed**, so `what-is-a-nameserver` can keep the line telling readers that people write it both ways.
- **"in this article" is banned only as an opener.** The anti-pattern is "In this article, we'll explore...". Nine `manage-*-record` articles say "the instructions in this article assume you are familiar with...", which is a fine sentence. A blanket match would have forced nine pointless rewrites.

## Deliberately not enforced

- **Contractions.** 487 existing uses across 112 articles, some deliberate. Decided not to enforce.
- **Vague link text** ("here", "this article"). 16 uses, each needing a rewritten phrase. Worth its own pass.

Both are recorded in a comment at the top of the test, so the omission is a decision rather than an oversight.

## Next, if this lands

The em-dash and contraction cleanup commits are the evidence that this category of work is being done by hand. CI extensions worth considering after this: anchors in cross-links (#2153), and imperative-form H1s on how-tos, which needs a way to tell a how-to from an explanation first.


dallasread (approved):
Sure! Why not! 😃

